PR #120, opened 2022-02-10
docs: make further weekly downloads numbers dynamic

brettz9:
I haven't confirmed this by attempting a build, but I think you want it dynamic like this as presumably one wishes the site to tout the newer 23.5 million over the currently listed 6.5 million.


nzakas (approved):
Nice catch! Thank you.

PR #122, opened 2022-02-11
fix: og:image

38elements:
`og:image` is url, not path.
https://ogp.me/
SVG is not supported at twitter card.
https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image


nzakas (approved):
Thank you!

PR #123, opened 2022-02-11
fix: focus outlines

SaraSoueidan:
- focus outline adjustments
- syntax highlighter colors matching those we have in the new docs

### DIFF
--- a/src/assets/css/styles.css
+++ b/src/assets/css/styles.css
@@ -363,24 +363,24 @@ h5:target,
 h6:target {
   background-color: var(--lighter-background-color); }
 
-*:focus {
-  outline: 3px solid transparent; }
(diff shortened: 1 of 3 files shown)


nzakas:
Sorry, I'm a bit lost...which input field are you referring to? I don't think we have any on the site currently?

nzakas:
Just double-checking: there are no input fields correct?

And is the syntax highlighting the same as the increased contrast on the docs site?

SaraSoueidan:
I meant the search input. But for whatever reason it still gets the focus outline, even though it's using the same styles as the rest of the interactive elements.

Re syntax highlighter: yep! I just copy-pasted the code form the docs to here.

nzakas (approved):

PR #127, opened 2022-02-15
fix: deleted unnecessary files

SaraSoueidan:
Deleted some redundant files across the repo that are not needed for the site.


nzakas:
It looks like the build failed on this but I’m not sure why. Need to dig in before merging.

nzakas (approved):

nzakas:
No idea why the build was failing but the changes look fine.

PR #129, opened 2022-02-20
feat: increase tappable area on mobile nav links

SaraSoueidan:
I noticed the navigation links don't take up the entire width on  mobile, so I made them so in order to increase the tappable area.


nzakas (approved):
LGTM

PR #130, opened 2022-02-21
fix: Accurate pie chart and updated budget

nzakas:
Makes the pie chart on the donate page accurate and updated budget numbers to reflect reality.


SaraSoueidan:
Screen Shot 2022-02-21 at 21 29 01

I'm guessing you edited the SVG file (?) because the text in the SVG now has the weird kerning above, which typically happens when the text inside an SVG gets positioned weirdly.

nzakas:
I didn’t. This was generated directly from Google Sheets. I’m not sure how Hayden generated the original one, but this was the closest I could get without manually tweaking things.

SaraSoueidan:
I exported the one from Figma and used it on the site. You think Hayden could update the original one in the Figma file? I can then export it again and replace this one.

nzakas:
I could, but that’s not a scalable long-term solution. We will go through this whenever the budget changes, so we need something that doesn’t involve manual labor.

Right now I’m of the mind that the Google Sheets export is good enough. The number of people who will look at it with any level of detail is vanishingly small and it’s more important to have an easy way to update than to worry too much about the letter spacing.

SaraSoueidan (approved):

PR #131, opened 2022-02-21
chore: More efficient donations fetch script

nzakas:
This just updates the script to fetch one-time donations to use the new Open Collective API for doing so. Instead of making multiple requests and filtering, we can now just request the one-time donations directly. :tada:


SaraSoueidan:
Heads up: Deploy seems to be failing.

nzakas:
Yeah, looks like the same random break we got on that other PR. Doesn’t appear related to this PR specifically.

PR #140, opened 2022-02-23
fix: Only load search.js when search input is there

nzakas:
We were loading `search.js` on all pages, but that caused an error because (unlike the docs site) we don't have a search box on every page.


SaraSoueidan (approved):
Good catch! Thank you.

PR #141, opened 2022-02-24
feat: Animation

cassieevans:
Kapture 2022-02-24 at 14 36 42

✨ Header animation

- disabled for older browsers
- no animation for users with prefers reduced motion settings enabled
- content visible without JS
- visually hidden title for screenreader users
- USing [GSAP](https://greensock.com/docs/) & [FLIP](https://greensock.com/docs/v3/Plugins/Flip) plugin


nzakas:
🔥 This looks great! The new lighter color on the dark theme makes all the difference.

Would you mind signing the electronic CLA (previous comment from the bot)? That just grants us a license to use your code in the project.

@SaraSoueidan any feedback?

SaraSoueidan:
This looks awesome! a11y is also good to go. ✨

I just noticed something: I got the darker popup on the dark theme on my first visit to the site. When I _manually_ changed the themes (using the theme switcher), _then_ I got the lighter popup.

nzakas:
Ooh good catch! I just noticed that too. 

It looks like the issue occurs when the OS preference is for dark theme. If I click on dark theme and then reload, then it works okay. 

I think maybe this line isn’t working correctly due to a race condition:
https://github.com/eslint/new.eslint.org/pull/141/files#diff-e2e6317a53d2812692c484ddeae950f9e496284d4da579cb3155c14c1f5c08a0R18

nzakas:
@cassieevans just wanted to check back to see if you can sign the CLA? Once you do, we can merge it and work through the slight bug with dark theme on our own. Thanks!

cassieevans:
I'm so sorry folks! I wasn't getting email updates for this for some reason!

I just got the email from Nicholas - I'll check this out today!

nzakas:
Thanks!

cassieevans:
Update - I found another little bit of behaviour that needs ironing out - the cursor follows you down the page if you scroll down at a certain point in the animation.

Just working out how to get around that. 🙃

cassieevans:
I think that's good to go! 🥳

Thanks for your patience folks.

nzakas:
This is looking 🔥 

It does look like the light/dark bug is fixed and I don’t see the cursor if the page is scrolled. 👍 

@SaraSoueidan do you want to take a quick look too?

nzakas (approved):
LGTM. Thanks so much for this!

PR #144, opened 2022-03-04
feat: make theme switcher more accessible to voice control users

SaraSoueidan:
- ✨ add visual labels to theme switcher buttons to make them more accessible to voice control users


nzakas (approved):
🎉 lgtm

PR #150, opened 2022-03-12
chore: fix layout shift

harish-sethuraman:
Avoids shifting content to the bottom when the ad pops up

Before:

https://user-images.githubusercontent.com/32865581/158005633-f1ff5ccd-1d90-457b-b747-357213d97591.mov

After:

https://user-images.githubusercontent.com/32865581/158005690-01cbb6c3-dc0e-4cf6-9a50-c244d496bd00.mov

### DIFF
--- a/src/assets/scss/components/carbon-ads.scss
+++ b/src/assets/scss/components/carbon-ads.scss
@@ -9,6 +9,13 @@
     padding: initial;
 }
 
+.carbon-wrapper {
+    height: 290px;
+    @media all and (max-width: 800px) {
+        display: none !important;
+    }
+}
+
 #carbonads {
     display: inline-block;
     margin: 2rem 0;;
--- a/src/content/pages/index.html
+++ b/src/content/pages/index.html
@@ -44,7 +44,9 @@ eleventyExcludeFromCollections: true
             </div>
         </div>
         <div class="span-10-12 eslint-versions-col">
-            {% include "partials/carbon-ad.html" %}
+            <div class="carbon-wrapper">
+                {% include "partials/carbon-ad.html" %}
+            </div>
             <dl class="eslint-versions" aria-labelledby="eslint-versions-title">
                 <span id="eslint-versions-title" hidden>{{ site.homepage.versions.title }}</span>
                 <dt>{{ site.homepage.versions.latest }}</dt>


snitin315 (commented):
Usually, we have Netlify deploy preview on PRs to quickly see but I don't see it on this PR. Maybe because it's coming from a fork.

/cc @nzakas

SaraSoueidan:
The ad is not supposed to be displayed on mobile AFAIK. @nzakas could you confirm? I have it explicitly set to `display: none` in the stylesheet. Did something change?

harish-sethuraman:
> The ad is not supposed to be displayed on mobile AFAIK. @nzakas could you confirm? I have it explicitly set to display: none in the stylesheet. Did something change?

Yes, We don't need the wrapper to take space in mobile right that is why added display none for that wrapper. The behaviour is the same. except we add explicit height in devices with ad and `display: none` the ad wrapper when ads are not needed (mobile)

SaraSoueidan:
I’m sorry you totally lost me here. x) The ads never displayed on mobile before. The CSS hid them, and even the script that adds the ad to the hero only runs at viewports > 800px. What am I missing?

harish-sethuraman:
```
<div class="carbon-wrapper">
            {% include "partials/carbon-ad.html" %}
</div>
```

The media query I've added is for the wrapper that I've introduced. It does not affect any of the existing styles.

nzakas:
Deploy previews only work for forks as a security precaution.

The ad isn't displayed on mobile, that's correct. I think what's happening here is that `carbon-wrapper` has a fixed height, and on mobile, that would show up as a big vertical gap in the page, so the media query ensures that it's hidden. Did I get that correct?

harish-sethuraman:
> Did I get that correct?
Yes exactly

harish-sethuraman:
Great thanks! 😄

snitin315 (approved):

snitin315:
For better understanding, see the `carbon ad` and layout shift below. 

## Before

https://user-images.githubusercontent.com/46647141/158918638-a3eea3c6-15cd-4ee9-a895-5e781ec63325.mov

## After

https://user-images.githubusercontent.com/46647141/158918691-16f87773-f1d0-440f-ad89-336b9e0bb6e1.mov

nzakas (approved):

nzakas:
Checked it out locally with mobile, too. Looks good. Thanks!

PR #151, opened 2022-03-12
fix: remove homepage horizontal overflow on mobile

adrienbaron:
This fixes the horizontal scrolling on mobile on the homepage. I reproduced on Safari mobile on iOS 15.3.1.

This was due to the cursor overflowing to the right. I fixed it by adding an `overflow-y: hidden` to the home page.

I'm happy to amend if you think this should be fixed differently 😊

**Before:**
image

**After:**
image


SaraSoueidan (approved):
LGTM.

PR #168, opened 2022-03-17
chore: added redirects

harish-sethuraman:
Addresses:
- https://github.com/eslint/new.eslint.org/issues/139

### DIFF
--- a/src/static/_redirects
+++ b/src/static/_redirects
@@ -1,0 +1,44 @@
+# Netlify Redirect Rules
+# https://www.netlify.com/docs/redirects/
+
+# Redirect default Netlify subdomain to primary domain
+https://eslint.netlify.com/* https://eslint.org/:splat 301!
+
+# External Redirects
+/cla                                https://cla.js.foundation/eslint/eslint 302!
+/conduct                            https://code-of-conduct.openjsf.org/ 302!
+/chat/tsc-meetings                  https://discord.gg/3brqTzJ 302!
+/chat/help                          https://discord.gg/nJdtBaF 302!
+/chat                               https://discord.gg/8szcydm 302!
+
+# Internal Redirects
+/docs                               https://eslint.org/ 301!
+/docs/user-guide/rules              https://eslint.org/docs/rules 301!
+/docs/rules/id-blacklist            https://eslint.org/docs/rules/id-denylist 301!
+
+/docs/0.24.1/command-line-interface https://eslint.org/docs/user-guide/command-line-interface 301!
+/docs/0.24.1/configuring            https://eslint.org/docs/user-guide/configuring 301!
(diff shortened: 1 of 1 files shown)


src/static/_redirects
```diff
@@ -0,0 +1,44 @@
...
+/chat/help                          https://discord.gg/nJdtBaF 302!
+/chat                               https://discord.gg/8szcydm 302!
+
+# Internal Redirects
+/docs                               https://eslint.org/ 301!
+/docs/user-guide/rules              https://eslint.org/docs/rules 301!
+/docs/rules/id-blacklist            https://eslint.org/docs/rules/id-denylist 301!
+
+/docs/0.24.1/command-line-interface https://eslint.org/docs/user-guide/command-line-interface 301!
+/docs/0.24.1/configuring            https://eslint.org/docs/user-guide/configuring 301!
```
snitin315, line 20: I believe we don't need these internal redirects for docs since we will have a separate website for docs.

/cc @nzakas can you confirm?

harish-sethuraman: > I believe we don't need these internal redirects for docs since we will have a separate website for docs.

Thought we might later on update the links. For now wanted to add proper working links

nzakas: For now all we need is /docs/* redirecting to eslint.org/docs/*

Later we can change to docs.eslint.org/*

harish-sethuraman: done

src/static/_redirects
```diff
@@ -12,33 +9,4 @@ https://eslint.netlify.com/* https://eslint.org/:splat 301!
...
-/docs/2.0.0/user-guide/rules        https://eslint.org/docs/rules 301!
-
-/docs/2.13.1/command-line-interface https://eslint.org/docs/user-guide/command-line-interface 301!
-/docs/2.13.1/configuring            https://eslint.org/docs/user-guide/configuring 301!
-/docs/2.13.1/integrations           https://eslint.org/docs/integrations 301!
-/docs/2.13.1/user-guide/rules       https://eslint.org/docs/rules 301!
-
-/docs/3.0.0/user-guide/rules        https://eslint.org/docs/rules 301!
\ No newline at end of file
+/docs/*                               https://eslint.org/ 301!
```
nzakas, line 12:
```suggestion
/docs/*                               https://eslint.org/docs/:splat 302!
```

We need to cover all docs URLs and definitely don’t want this to be permanent.

nzakas (commented):
Just need to be sure the docs go to the correct location.

src/static/_redirects
harish-sethuraman: Done

snitin315 (approved):
LGTM, thanks for contributing.

nzakas (approved):
LGTM. Thanks!

PR #178, opened 2022-03-25
chore: Fix es.eslint.org build by upgrading binary deps

btmills:
The es.eslint.org build is failing due to a `node-sass` binary installation issue. [`node-sass` is deprecated](https://sass-lang.com/blog/libsass-is-deprecated), so this is not entirely unsurprising. We already have its replacement, `sass`, installed, so I uninstalled `node-sass` and updated package scripts to use `sass`.

The new `sass` implementation didn't allow top-level CSS custom properties inside the `@media` query, so I nested them inside a `:root` node like the other custom properties in the file.

I took a similar approach of upgrading `jpegtran-bin` to get the build working.


btmills:
1. Since we're committing both SCSS and the generated CSS, I included the changes from `npm run build` in the first commit. Would we prefer to generate those on the fly during build and not commit them?
1. To minimize the size of the change to `package-lock.json` in the second commit, is there a better way of upgrading just a transitive dependency than `npm uninstall imagemin imagemin-cli && npm install -D imagemin imagemin-cli`? In yarn, I could remove the lockfile entry for only that dependency and reinstall.

nzakas:
My hero! 😍 Still doesn’t seem like one build should fail while the other passes, but working around it works too. 

> Would we prefer to generate those on the fly during build and not commit them?

Yes, that’s been on my todo list to look at. Not a big deal but I’d be happy if that got cleaned up.

I’m not familiar enough with lock files to know much about optimizing them.

nzakas (approved):

PR #182, opened 2022-03-26
fix: robots.txt URL

nzakas:
The robots.txt file needs a fully-qualified URL. We were generating an invalid robots.txt file because `metadata.url` was not defined.


nzakas:
The failures here were rate limits, which will be fixed by #181

PR #188, opened 2022-04-13
fix: Ensure search only searches blog posts

nzakas:
We need to filter the Algolia search so it only shows blog posts on the blog index page. I'm pretty sure this will do the trick (based on [the docs](https://www.algolia.com/doc/api-reference/api-parameters/filters/#examples), but want to double-check with deploy preview.

### DIFF
--- a/src/assets/js/search.js
+++ b/src/assets/js/search.js
@@ -34,7 +34,7 @@ const searchClearBtn = document.querySelector('#search__clear-btn');
  */
 function fetchSearchResults(query) {
     return index.search(query, {
-        // facetFilters: ["tags:docs"]
+        filters: "blog"
     }).then(({ hits }) => hits);
 }
 
--- a/src/assets/js/search.js
+++ b/src/assets/js/search.js
@@ -71,8 +71,8 @@ function displaySearchResults(results) {
             const listItem = document.createElement('li');
             listItem.classList.add('search-results__item');
             listItem.innerHTML = `
-                <h2 class="search-results__item__title"><a href="${result.url}">${result.hierarchy.lvl0}</a></h2>
-                <p class="search-results__item__context">${result._highlightResult.hierarchy.lvl0.value}</p>
+                <h2 class="search-results__item__title"><a href="${result.url}">${result.hierarchy.lvl1}</a></h2>
+                <p class="search-results__item__context">${result._highlightResult.hierarchy.lvl1.value}</p>
             `.trim();
             list.append(listItem);
         }
--- a/src/assets/js/search.js
+++ b/src/assets/js/search.js
@@ -34,7 +34,7 @@ const searchClearBtn = document.querySelector('#search__clear-btn');
  */
 function fetchSearchResults(query) {
     return index.search(query, {
-        filters: "blog"
+        facetFilters: ["tags:blog"]
     }).then(({ hits }) => hits);
 }
 


nzakas:
Hmmm, this didn't work. Now we're getting no results instead of filtering on the blog posts. :(

src/assets/js/search.js
```diff
@@ -34,7 +34,7 @@ const searchClearBtn = document.querySelector('#search__clear-btn');
  */
 function fetchSearchResults(query) {
     return index.search(query, {
-        // facetFilters: ["tags:docs"]
+        filters: "blog"
```
nzakas, line 37: We definitely have content marked with `tag:blog` and `tag:docs`. We are using this in the current site:

https://github.com/eslint/website/blob/master/src/js/main/index.js#L28

kfranqueiro:
I'm not really sure how this can end up working in one page and not another if it's using the same index with the same configuration (or are there two different `eslint` indexes under different apps/accounts?).

If I add `attributesToRetrieve: "*"` to the search options, I indeed see `tags` with values like `docs` and `blog`, but that doesn't seem to be working in `search.js` via the same `facetFilters` as in the file you linked to, nor via the `filters` equivalent which I believe would be `filters: "tags:blog"`. This leads me to wonder if the `tags` attribute [needs to be set up for faceting](https://www.algolia.com/doc/guides/managing-results/refine-results/faceting/how-to/declaring-attributes-for-faceting/) (or at least filtering) in the index configuration and currently is not?

src/assets/js/search.js
shortcuts: Hey, what you have on the current implementation (`algoliaOptions: { facetFilters: [["tags:docs", "tags:blog"]] },`), will behave the same as no `algoliaOptions`, as you only have two tags set in your index and nested arrays will behave as an `OR`.

You can actually keep the same logic but with the `blog` tags.

```suggestion
        facetFilters: ["tags:blog"]
```

Small sandbox with examples: https://codesandbox.io/s/new-frost-7jcxy6?file=/src/index.js

kfranqueiro: I would note that index.js and shortcuts' example use a different app ID than the code in search.js, which goes back to the question I raised - is this a tale of 2 different indexes, one of which is configured correctly while the other isn't?

shortcuts: DocSearch defaulted to `BH4D9OD16A` previously, which was a shared app. With our new infra, personal apps are now available and the ESLint one is `L633P0C2IR`

nzakas:
@shortcuts unfortunately, the change didn’t work (you can see in the deploy preview - no results returned for any query). I had tried that initially, too.

Could @kfranqueiro’s analysis be correct?

shortcuts:
> @shortcuts unfortunately, the change didn’t work (you can see in the deploy preview - no results returned for any query). I had tried that initially, too.
> 
> Could @kfranqueiro’s analysis be correct?

Looking at your index from the Crawler UI, I don't see the `tags` being set. You can either delete the index and restart a crawl or directly add `tags` from this page: https://www.algolia.com/apps/L633P0C2IR/explorer/configuration/eslint/facets

Screenshot 2022-04-14 at 20 35 36

nzakas:
Ah interesting. I deleted the index and recrawled, and it seems to have worked! Just pushed a small tweak to what is displayed in the dropdown so we can verify on the deploy preview. Thanks everyone!

PR #194, opened 2022-04-19
fix: fix color of selection in copy

harish-sethuraman:
Fixes 
- https://github.com/eslint/new.eslint.org/issues/190


nzakas:
Just enabled deploy preview so we can take a look. I’m on mobile right now so the styles don’t apply.

snitin315 (approved):
LGTM, thanks!

nzakas (approved):

PR #196, opened 2022-04-20
fix: theme flashing

harish-sethuraman:
When the site loads and no theme is set in the localstorage the site flashes between dark and light mode (when the system is in dark mode)

### How to reproduce: 
- set system theme as dark mode and clear the site's localstorage
- reload the page the page might flash 

Removed some code that are same as the one we already have inside layout
- https://github.com/eslint/new.eslint.org/blob/824c4bf34390a5781dc69f2ca993a873e37d4e15/src/_includes/layouts/base.html#L67-L74

Before: 

https://user-images.githubusercontent.com/32865581/164293001-dac07d15-1a86-4837-8205-a5257f6e42e7.mov

### DIFF
--- a/src/_includes/layouts/base.html
+++ b/src/_includes/layouts/base.html
@@ -65,10 +65,12 @@
     <link rel="preconnect" href="https://www.googletagmanager.com/">
 
     <script>
-            let theme = window.localStorage.getItem("theme");
+        (function() {
+            var theme = window.localStorage.getItem("theme");
             if (!theme && window.matchMedia('(prefers-color-scheme: dark)').matches) {
                 document.documentElement.setAttribute('data-theme', 'dark');
             } else if (theme) document.documentElement.setAttribute('data-theme', theme);
+        })();
     </script>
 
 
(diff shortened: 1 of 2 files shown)


snitin315 (commented):
Looks good to me.

/cc @SaraSoueidan do you want to take a look at this?

src/_includes/layouts/base.html
```diff
@@ -65,12 +65,10 @@
     <link rel="preconnect" href="https://www.googletagmanager.com/">
 
     <script>
-        (function() {
```
nzakas, line 68: Seems like we should keep this to avoid creating a global? (Same variable name as in theme.js)

nzakas (changes requested):

src/_includes/layouts/base.html
harish-sethuraman: Agreed

harish-sethuraman: Removed thanks :)

SaraSoueidan:
Good catch! LGTM!

nzakas (approved):

PR #208, opened 2022-05-07
fix: set default theme

amareshsm:
**Current behavior:** 
Theme not set initially both light & dark theme animated dropdowns are shown. Below I have attached screenshots for your reference.

| In desktop  | In mobile |
| ------------- | ------------- |
| image  |  image  |  

**Steps to reproduce:**
When we open the website for the first time this bug will be reproduced when the theme is not selected on your device or when then the theme is light.
Device: MacBook Pro 2019 
OS: macOS Big Sur 
Browser: Chrome Version 100.0.4896.60
Ref: https://share.vidyard.com/watch/At5DoYpzWTnJTX6JzqTxxs

**Change:**
Set system prefers-color-scheme as the initial theme.

### DIFF
--- a/src/assets/js/themes.js
+++ b/src/assets/js/themes.js
@@ -1,31 +1,27 @@
 /* theme toggle buttons */
-(function() {
-    const enableToggle =btn=> {
+(function () {
+    const enableToggle = btn => {
         btn.setAttribute("aria-pressed", "true");
     }
-
-    const disableToggle =btn => {
+    const disableToggle = btn => {
         btn.setAttribute("aria-pressed", "false");
     }
-    const setTheme = theme=>{
+    const setTheme = theme => {
         document.documentElement.setAttribute('data-theme', theme);
         window.localStorage.setItem("theme", theme);
     }
 
     let theme = window.localStorage.getItem("theme");
-    if(!theme || !(theme==="light") || !(theme==="dark")){
+    if (!theme || !(theme === "light") || !(theme === "dark")) {
         let isDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
         theme = isDark ? "dark" : "light";
     }
 
-    document.addEventListener('DOMContentLoaded', function() {
+    document.addEventListener('DOMContentLoaded', function () {
         const switcher = document.getElementById('js-theme-switcher');
         switcher.removeAttribute('hidden');
-
         const light_theme_toggle = document.getElementById('light-theme-toggle'),
             dark_theme_toggle = document.getElementById('dark-theme-toggle');
-
-
         if (theme === "light") {
             enableToggle(light_theme_toggle);
             disableToggle(dark_theme_toggle);
@@ -35,21 +31,18 @@
             disableToggle(light_theme_toggle);
             setTheme(theme);
         }
-
-        light_theme_toggle.addEventListener("click", function() {
+        light_theme_toggle.addEventListener("click", function () {
             enableToggle(light_theme_toggle);
             theme = this.getAttribute('data-theme');
             setTheme(theme);
             disableToggle(dark_theme_toggle);
-            
-        }, false);
 
-        dark_theme_toggle.addEventListener("click", function() {
+        }, false);
+        dark_theme_toggle.addEventListener("click", function () {
             enableToggle(dark_theme_toggle);
             theme = this.getAttribute('data-theme');
             setTheme(theme);
             disableToggle(light_theme_toggle);
         }, false);
     }, false);
-
 })();
--- a/src/assets/js/themes.js
+++ b/src/assets/js/themes.js
@@ -13,6 +13,7 @@
 
     var theme = window.localStorage.getItem("theme");
     if (!theme || (theme !== "light" && theme !== "dark")) {
+        var isChanged = true;
         var isDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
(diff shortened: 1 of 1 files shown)


src/assets/js/themes.js
```diff
@@ -1,49 +1,48 @@
 /* theme toggle buttons */
-(function() {
-    var enableToggle = function(btn) {
+(function () {
+    const enableToggle = btn => {
```
aladdin-add, line 3: we were using `var` and `iife` (ES5-only) here, as the js won't be compiled? I'm not sure, but it could have a reason. :)

aladdin-add (commented):
the change LGTM, just a question needs someone to confirm. thanks!

src/assets/js/themes.js
amareshsm: converted to ES5

src/assets/js/themes.js
```diff
@@ -1,49 +1,48 @@
...
+    var disableToggle = function (btn) {
         btn.setAttribute("aria-pressed", "false");
     }
+    var setTheme = function (theme) {
+        document.documentElement.setAttribute('data-theme', theme);
+        window.localStorage.setItem("theme", theme);
+    }
 
+    var theme = window.localStorage.getItem("theme");
+    if (!theme || (theme !== "light" && theme !== "dark")) {
```
harish-sethuraman, line 15:
```suggestion
    if (!theme) {
```

When the theme value is falsy we are not going to have it as `light` and `dark`?  I think we don't need those conditions?

src/assets/js/themes.js
```diff
@@ -1,49 +1,48 @@
...
         var light_theme_toggle = document.getElementById('light-theme-toggle'),
             dark_theme_toggle = document.getElementById('dark-theme-toggle');
-
-
-        if (theme == "light") {
+        if (theme === "light") {
             enableToggle(light_theme_toggle);
             disableToggle(dark_theme_toggle);
-        } else if (theme == "dark") {
+            setTheme(theme);
```
harish-sethuraman, line 28: I think we don't need this part of the code too... 
- https://github.com/eslint/new.eslint.org/blob/7bd0f8f4c3be64337f2392203cf952d8a512143b/src/_includes/layouts/base.html#L68-L73

During initial page load this method actually setstheme as dark mode if the system preference is dark else we dont set any theme.

src/assets/js/themes.js
```diff
@@ -1,49 +1,48 @@
...
+        } else if (theme === "dark") {
             enableToggle(dark_theme_toggle);
             disableToggle(light_theme_toggle);
+            setTheme(theme);
```
harish-sethuraman, line 32: I think we don't need this part of the code too... 
- https://github.com/eslint/new.eslint.org/blob/7bd0f8f4c3be64337f2392203cf952d8a512143b/src/_includes/layouts/base.html#L68-L73

During initial page load this method actually setstheme as dark mode if the system preference is dark else we dont set any theme.

src/assets/js/themes.js
amareshsm: yes if the theme value is falsy it won't be either light or dark. Suppose the theme value is truthy then we are making sure that it must be either light or dark. Just an additional check only. In case the local store value accidentally changed to something else other than light or dark.

src/assets/js/themes.js
amareshsm: Yes, the above code works perfectly if the user preference is dark but if not then we are not setting any theme. 

https://github.com/eslint/new.eslint.org/blob/7bd0f8f4c3be64337f2392203cf952d8a512143b/src/assets/scss/animation.scss#L104-L114

In styles using theme only, we are deciding which dropdown (light or dark) should be shown. So the above block of code is not working as expected.

One way we can set the theme to light is if the system preference is not dark.
```
    if (!theme && window.matchMedia('(prefers-color-scheme: dark)').matches) { 
	         document.documentElement.setAttribute('data-theme', 'dark'); 
	} else
	  document.documentElement.setAttribute('data-theme', 'light'); //like this
```
But if we do like this it will break the theme toggle logic. It will always be based on the system's preference.

So in the theme.js I am initially setting the theme based on preference and updating it in the local store. 
If the theme is set once the below logic will set the theme attribute based on the local store data.
https://github.com/eslint/new.eslint.org/blob/7bd0f8f4c3be64337f2392203cf952d8a512143b/src/assets/scss/animation.scss#L104-L114

And manual toggling using the toggle button in UI will be handled by the below code.

https://github.com/eslint/new.eslint.org/blob/cc53724d80a2200a724e3756c38e7e06b1482fc1/src/assets/js/themes.js#L34-L46


Correct me if I am wrong and open to suggestions/comments.

harish-sethuraman: Yea I get it :) I just thought it was better to have the initial theme setting done in the base.html instead of doing similar stuff at `base.html` as well as `theme.js`. But anything is fine with me too :)

aladdin-add (approved):

nzakas:
Can you please provide some steps to reproduce? And what browser and OS are you using?

amareshsm:
> Can you please provide some steps to reproduce? And what browser and OS are you using?

I have updated the PR. Added steps to reproduce.

nzakas (approved):
Thanks, I was able to reproduce locally and confirmed that this change fixes the issue. Nice catch and thanks for your help

PR #209, opened 2022-05-08
chore: replace addListener with addEventListener

amareshsm:
use addEventListener() instead of matchMedia().addListener() since matchMedia().addListener() is deprecated.


nzakas (approved):
LGTM. Thanks.

PR #211, opened 2022-05-10
fix: removes layout shift when opening navbar in mobile

harish-sethuraman:
IDK if this was intentional but thought it would be better if there was no shift when opening nav

Before:

https://user-images.githubusercontent.com/32865581/167675201-4dd8a930-4073-40d2-b533-81cdf381e90f.mov

After:

https://user-images.githubusercontent.com/32865581/167675228-4a8d88fb-930c-45c0-967d-50ec97c42036.mov

### DIFF
--- a/src/assets/scss/animation.scss
+++ b/src/assets/scss/animation.scss
@@ -15,7 +15,7 @@
       height: 100%;
       transform: scaleX(var(--scale));
       transform-origin: center right;
-      z-index: 999;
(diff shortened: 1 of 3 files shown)


nzakas:
@SaraSoueidan do you recall if this was intentional?

nzakas:
Confirmed that this behavior is intentional.

PR #212, opened 2022-05-11
fix: highlight selected theme

amareshsm:
### Description:
Highlighted selected theme.

### Before:
User selected theme was not highlighted. It is a miss from my end in my previous PR [#208](https://github.com/eslint/new.eslint.org/pull/208) 

### Steps to reproduce:
Select a theme and reload the page. 

### After:
Highlighted the user selected theme.

### Details:
**base.html** 
- Gives first preference to user selected theme. If the theme is not available apply based on system preferences. 

**theme.js** 
-  highlights the selected theme. 
-  Toggles the theme when it is changed and stores it.

### DIFF
--- a/src/_includes/layouts/base.html
+++ b/src/_includes/layouts/base.html
@@ -65,12 +65,13 @@
     <link rel="preconnect" href="https://www.googletagmanager.com/">
 
     <script>
-        (function() {
-            var theme = window.localStorage.getItem("theme");
-            if (!theme && window.matchMedia('(prefers-color-scheme: dark)').matches) {
-                document.documentElement.setAttribute('data-theme', 'dark');
-            } else if (theme) document.documentElement.setAttribute('data-theme', theme);
-        })();
+            (function () {
+                var theme = window.localStorage.getItem("theme");
+                if (theme) document.documentElement.setAttribute('data-theme', theme)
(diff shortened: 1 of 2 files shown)


src/_includes/layouts/base.html
```diff
@@ -65,12 +65,13 @@
...
-        (function() {
-            var theme = window.localStorage.getItem("theme");
-            if (!theme && window.matchMedia('(prefers-color-scheme: dark)').matches) {
-                document.documentElement.setAttribute('data-theme', 'dark');
-            } else if (theme) document.documentElement.setAttribute('data-theme', theme);
-        })();
+            (function () {
+                var theme = window.localStorage.getItem("theme");
+                if (theme) document.documentElement.setAttribute('data-theme', theme)
+                else if (!theme && window.matchMedia('(prefers-color-scheme: dark)').matches)
```
harish-sethuraman, line 71: We have already filtered if theme is true in the first condition and we can expect it to be false in the next all conditions ?
```suggestion
                else if (window.matchMedia('(prefers-color-scheme: dark)').matches)
```

amareshsm: yes we can remove the theme check.

nzakas (approved):
LGTM. Thanks!

PR #214, opened 2022-05-13
fix: added missing logos in blogs

harish-sethuraman:
Check this page and seems like logos are missing: https://new.eslint.org/blog/2021/12/contra-gold-sponsor-eslint/

That page is the copy of https://eslint.org/blog/2021/12/contra-gold-sponsor-eslint

So used logos from that page in this blog. 

PS: The class name like `text-center` is not available in the `blog.scss`


nzakas:
Thanks. We are actually creating new images for the new blog post format. Some of them are already up, but for some reason, not being referenced. I think we had a bit of a copy-paste issue between the old blog and the new one. Here's an example of the new image format:
https://github.com/eslint/new.eslint.org/blob/main/src/assets/images/blog-covers/sponsorship-chrome.png

I just need to go back, create the remaining images, and update the blog posts.

harish-sethuraman:
> I just need to go back, create the remaining images, and update the blog posts.

Are these images being created manually? for every sponsor?

nzakas:
Sadly yes. I’d like to figure out a way to automate it but haven’t had the time.

PR #215, opened 2022-05-13
perf: use system's own flag on non-Windows systems

kecrily:
This is an addition to https://github.com/eslint/new.eslint.org/pull/213

### DIFF
--- a/src/assets/scss/tokens/typography.scss
+++ b/src/assets/scss/tokens/typography.scss
@@ -79,5 +79,5 @@
 }
 
 .win {
-    font-family: "Twemoji Country Flags"!important;
+    font-family: "Twemoji Country Flags" !important;
 }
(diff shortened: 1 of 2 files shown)


src/assets/scss/tokens/typography.scss
```diff
@@ -78,3 +77,7 @@
             "Segoe UI Emoji",
             "Segoe UI Symbol";
 }
+
+.win {
+    font-family: "Twemoji Country Flags";
```
nzakas, line 82: Do we really want to set this as the font family for the whole page?

kecrily: Now I've changed it to use only in the language selector

src/assets/scss/tokens/typography.scss
```diff
@@ -79,5 +79,5 @@
 }
 
 .win {
-    font-family: "Twemoji Country Flags";
+    font-family: "Twemoji Country Flags"!important;
```
nzakas, line 82:
```suggestion
    font-family: "Twemoji Country Flags" !important;
```

nzakas (approved):

nzakas:
Thanks!

PR #223, opened 2022-05-31
build: compress build css and remove unwanted parallel

harish-sethuraman:
use `watch:sass` instead of `build:sass` and `npm-run-all` requires only one `--parallel`

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
         "images": "imagemin '_site/assets/images' --out-dir='_site/assets/images'",
         "watch:sass": "sass --watch src/assets/scss:src/assets/css",
         "watch:eleventy": "eleventy --serve --port=2022",
-        "build:sass": "sass src/assets/scss:src/assets/css",
+        "build:sass": "sass --no-source-map --style=compressed src/assets/scss:src/assets/css",
         "build:eleventy": "npx @11ty/eleventy",
         "fetch:team": "node tools/fetch-team-data.js",
         "fetch:stats": "node tools/fetch-stats.js",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
         "fetch:sponsors": "node tools/fetch-sponsors.js",
         "fetch": "npm-run-all --parallel fetch:*",
         "start": "npm-run-all --parallel watch:* images",
-        "build": "npm-run-all --parallel build:* images"
+        "build": "npm-run-all build:* images"
     },
     "devDependencies": {
         "@11ty/eleventy": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
         "fetch:sponsors": "node tools/fetch-sponsors.js",
         "fetch": "npm-run-all --parallel fetch:*",
         "start": "npm-run-all --parallel watch:* images",
-        "build": "npm-run-all build:* images"
+        "build": "npm-run-all build:sass build:eleventy images"
     },
     "devDependencies": {
         "@11ty/eleventy": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -19,8 +19,8 @@
         "fetch:stats": "node tools/fetch-stats.js",
         "fetch:sponsors": "node tools/fetch-sponsors.js",
         "fetch": "npm-run-all --parallel fetch:*",
-        "start": "npm-run-all build:sass --parallel watch:*  --parallel images",
-        "build": "npm-run-all build:sass --parallel build:*   --parallel images"
+        "start": "npm-run-all --parallel watch:* images",
+        "build": "npm-run-all --parallel build:* images"
     },
     "devDependencies": {
         "@11ty/eleventy": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
         "fetch:sponsors": "node tools/fetch-sponsors.js",
         "fetch": "npm-run-all --parallel fetch:*",
         "start": "npm-run-all build:sass --parallel watch:* images",
-        "build": "npm-run-all build:sass --parallel build:eleventy images"
+        "build": "npm-run-all build:sass build:eleventy images"
     },
     "devDependencies": {
         "@11ty/eleventy": "^1.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -19,8 +19,8 @@
         "fetch:stats": "node tools/fetch-stats.js",
         "fetch:sponsors": "node tools/fetch-sponsors.js",
         "fetch": "npm-run-all --parallel fetch:*",
-        "start": "npm-run-all build:sass --parallel watch:*  --parallel images",
-        "build": "npm-run-all build:sass --parallel build:*   --parallel images"
+        "start": "npm-run-all --parallel watch:* images",
+        "build": "npm-run-all --parallel build:* images"
```
nzakas, line 23: I’m pretty sure we are intentionally running build:sass first and then the rest in parallel

harish-sethuraman: But? `build:sass` is run first and then again `watch:sass` is run in parallel again since we use `watch:*`?

harish-sethuraman: Also in the `build` command the `sass` is built first and then again it is being built in parallel? when using `build:*`

mdjermanovic (changes requested):

package.json
```diff
@@ -19,8 +19,8 @@
         "fetch:stats": "node tools/fetch-stats.js",
         "fetch:sponsors": "node tools/fetch-sponsors.js",
         "fetch": "npm-run-all --parallel fetch:*",
-        "start": "npm-run-all build:sass --parallel watch:*  --parallel images",
-        "build": "npm-run-all build:sass --parallel build:*   --parallel images"
+        "start": "npm-run-all --parallel watch:* images",
+        "build": "npm-run-all build:* images"
```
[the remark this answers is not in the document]
harish-sethuraman: npm-run-all executes commands in the order in which they are present so it executes sass first and then eleventy. However If needed I can change it to your suggestion if needed? 

```
$ npm-run-all build:* images
$ sass src/assets/scss:src/assets/css
$ npx @11ty/eleventy
```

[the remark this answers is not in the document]
mdjermanovic: I would rather specify the order directly in the `build` script than remember that the order of keys in `"scripts": {}` is important.

package.json
nzakas: For watch: I think this is because we need an initial sass build for the other builds to work, then watch:sass watches for changes.

For build: I think we still need sass built first and it’s run a second time as a consequence of the wildcard rather than intentional.

harish-sethuraman: `other builds to work` but new.eslint.org doesn't pull any other bundles right? we won't need `build:sass`. even playground doesn't have this? Also how ever watch command is going to do a initial build (generates css) and then watch changes? 

- https://github.com/eslint/playground/blob/7a052c4fcbdd97e866fbf2bbe3f3e8e52289cd9a/tools/dev-server.js#L153-L155

For build: I've fixed it to run sass first and then eleventy

nzakas: What I mean is we need to build sass before we build Eleventy. The CSS files get placed into assets/css, so then Eleventy can pick those up when the site is generated.

The same is true for watching. Otherwise, we don’t know which task is run first, so the CSS files may not be in place when Eleventy is first built.

harish-sethuraman: I've added the `build:sass` to the start command and remove unwanted parallel command + stopped running `build:sass` twice

nzakas (approved):
LGTM. Just waiting for @mdjermanovic to verify his feedback.

package.json
```diff
@@ -13,14 +13,14 @@
...
-        "build:sass": "sass src/assets/scss:src/assets/css",
+        "build:sass": "sass --no-source-map --style=compressed src/assets/scss:src/assets/css",
         "build:eleventy": "npx @11ty/eleventy",
         "fetch:team": "node tools/fetch-team-data.js",
         "fetch:stats": "node tools/fetch-stats.js",
         "fetch:sponsors": "node tools/fetch-sponsors.js",
         "fetch": "npm-run-all --parallel fetch:*",
-        "start": "npm-run-all build:sass --parallel watch:*  --parallel images",
-        "build": "npm-run-all build:sass --parallel build:*   --parallel images"
+        "start": "npm-run-all build:sass --parallel watch:* images",
```
mdjermanovic, line 22: If we run `images` in parallel with eleventy, it will probably do nothing because it operates on a directory that eleventy has yet to copy from src.

I'm not sure if it was doing anything in the previous version as well. It looks like it wasn't run at all, and even if it is then it's after we finished watching. Maybe we should just remove it?

package.json
```diff
@@ -13,14 +13,14 @@
...
+        "build": "npm-run-all build:sass --parallel build:eleventy images"
```
mdjermanovic, line 23:
```suggestion
        "build": "npm-run-all build:sass build:eleventy images"
```

None of these 3 tasks can be performed in parallel with any other because each of them operates on the result of the previous task.

`build:eleventy` must wait for `build:sass` to compile `src/assets/scss/*` to `src/assets/css/*`

`images` must wait for `build:eleventy` to copy images to `_site/assets/images/`

mdjermanovic (changes requested):

package.json
harish-sethuraman: "start": "npm-run-all build:sass images --parallel watch:*",

May be we should do something like this. Run the sass build and then images build and start watching in parallel?

mdjermanovic: `images` works on `_site` which is initially empty so running `images` at that point would do nothing?

nzakas (approved):
LGTM. Just waiting for @mdjermanovic to verify.

package.json
harish-sethuraman: I was thinking if a new image was added during the watch mode the image command would be run only once? Should we may be consider removing that in the watch mode since it is run only once? or do it once the parallel commands execution was completed? like this
```suggestion
        "start": "npm-run-all build:sass --parallel watch:* --serial images",
```

mdjermanovic: Then the minified images won't be used in the local preview, and Eleventy will overwrite them on subsequent `npm start`. I think we just can't incorporate `images` in the watch mode, because it can be run either too early or too late, and it doesn't have its own watch mode. In my opinion, we should remove `images` from this script because it has no effect.

harish-sethuraman: Done!

mdjermanovic (approved):
LGTM, thanks!

PR #167, opened 2022-03-17
chore: Fixes layout shift in mobile caused due to topnav

harish-sethuraman:
Lighthouse scores improved as well

Before:

https://user-images.githubusercontent.com/32865581/158770541-db2818f9-9b66-4ae9-a289-ec22d84b96b9.mov

After:

https://user-images.githubusercontent.com/32865581/158770834-482c4560-0f68-4d26-aa31-dbdd66199d0d.mov

No shifts at all

PR #225, opened 2022-06-01
fix: show navlist when script disabled

amareshsm:
**Description:**
Show navbar elements when the script is disabled and reduce cls

Before:
Navbar list elements were not shown when js was disabled.
https://github.com/eslint/new.eslint.org/pull/167 reduces layout shift in mobile view but hiding navlist by default causes issue when script is disabled.  
image

After:
image

### DIFF
--- a/src/_includes/components/navigation.html
+++ b/src/_includes/components/navigation.html
@@ -1,7 +1,7 @@
 <nav class="site-nav" aria-label="Main">
     <div class="flexer">
         <a href="{{ links.donate }}" class="c-btn c-btn--primary donate-link">{{ site.donate_page.title }}</a>
-        <button class="site-nav-toggle" aria-label="Menu" id="nav-toggle" hidden>
+        <button class="site-nav-toggle" aria-label="Menu" id="nav-toggle">
             <svg width="20" height="20" viewBox="20 20 60 60">
                 <path id="ham-top"    d="M30,37 L70,37 Z" stroke="currentColor"></path>
                 <path id="ham-middle" d="M30,50 L70,50 Z" stroke="currentColor"></path>
@@ -9,7 +9,7 @@
             </svg>
         </button>
     </div>
-    <ul id="nav-list" data-open="false">
+    <ul id="nav-list">
         {% for item in site.navigation %}
         <li>
             {% set itemLink = links[item.link] %}
--- a/src/_includes/layouts/base.html
+++ b/src/_includes/layouts/base.html
@@ -63,6 +63,9 @@
         .no-js #nav-toggle {
             display: none !important;
         }
+        .no-js #copyBtn{
+            display: none !important;
+        }
         @media all and (min-width: 680px) {
             .enhanced #nav-toggle{
                 display: none;
@@ -71,6 +74,9 @@
         @media all and (max-width: 679px) {
             .no-js #nav-list {
                 display: block !important;
+            }
+            .enhanced #nav-list{
+                display: none;
             }
         }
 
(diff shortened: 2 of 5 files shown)


nzakas:
Can we just revert #167? This PR seems to have a lot of other unrelated changes.

amareshsm:
> Can we just revert #167? This PR seems to have a lot of other unrelated changes.

Reverting #167 will fix the issue but we still have [CLS in mobile view](https://github.com/eslint/playground/pull/64#discussion_r889605562).

src/_includes/components/navigation.html
```diff
@@ -1,15 +1,15 @@
 <nav class="site-nav" aria-label="Main">
     <div class="flexer">
         <a href="{{ links.donate }}" class="c-btn c-btn--primary donate-link">{{ site.donate_page.title }}</a>
-        <button class="site-nav-toggle" aria-label="Menu" id="nav-toggle" hidden>
```
amareshsm, line 4: Ref: https://github.com/eslint/playground/pull/64#discussion_r889601233

src/content/pages/index.html
```diff
@@ -29,7 +29,7 @@ <h1 class="section-title">{{ site.homepage.title }}</h1>
                 <div class="eslint-install-code-wrapper">
                     <div class="eslint-install-code" role="region" aria-labelledby="eslint-installation">
                         <span hidden id="eslint-installation">{{ site.homepage.install.title }}</span>
-                        <button class="eslint-install-code__btn" id="copyBtn" aria-labelledby="copy-button-label" hidden>
```
amareshsm, line 32: Removed hidden attribute to avoid layout shift.



Before:

https://user-images.githubusercontent.com/30730124/172023126-069a6750-748c-426f-b7a3-c244c2d1e29e.mov

After:


https://user-images.githubusercontent.com/30730124/172023139-ae91ee8c-82f1-48a1-9f4b-eab422a6a89a.mov

src/_includes/layouts/base.html
```diff
@@ -52,12 +52,35 @@
...
 
-
-    <script type="module">
+    <script>
         // This is a capable browser, let's improve the UI further!
         document.documentElement.classList.add("enhanced");
         document.documentElement.classList.remove('no-js');
     </script>
+ 
+    <style>
```
nzakas, line 61: Can this be moved into a .scss file?

Also, what if we made a `js-only` class that we can apply to all the buttons that should be hidden when there JS is disabled? That way, we wouldn't need to list out each element individually.

nzakas (commented):
Thanks for looking at this. I had a few questions and comments.

src/_includes/layouts/base.html
amareshsm: Sure, I will try it.

amareshsm: yes, changes are done.

src/_includes/layouts/base.html
```diff
@@ -52,12 +52,35 @@
     <meta name="twitter:image" content="{{ cover_image }}">
     <meta name="twitter:creator" content="@geteslint">
 
-
-    <script type="module">
+    <script>
```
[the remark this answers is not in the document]
amareshsm: if we keep this as a module that script will be deferred (`'enhanced'` class is getting added after parsing is completed) which introduces a layout shift.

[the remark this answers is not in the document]
nzakas: Ah okay, sounds good.

nzakas (approved):
LGTM. Thanks

PR #249, opened 2022-06-27
fix: Ensure landing page h1 element is exposed to assistive technologies

BenDMyers:
Currently, the eslint.org landing page's "Find and fix problems in your JavaScript code" `<h1>` is not exposed at all to assistive technologies such as screenreaders. From what I can tell, this seems to be out of a desire to not burden screenreader users with the red-squiggly-fixing animation, but it means that screenreader users don't get access to the heading contents at all.

To remedy this, I've removed the `aria-hidden` from the `<h1>`, and instead placed it on each of the animated child nodes. Additionally, I've added a visually-hidden span which contains the full, unbroken text contents of the heading.

### DIFF
--- a/src/content/pages/index.html
+++ b/src/content/pages/index.html
@@ -17,7 +17,7 @@ eleventyExcludeFromCollections: true
         <div class="span-1-7 content-container">
             <div class="">
                 <div class="section-head">
-			{% if site.homepage.animation.enabled %}
+		{% if site.homepage.animation.enabled %}
                     	<h1 class="visually-hidden">{{ site.homepage.title }}</h1>
                         {% include 'components/header-animation.njk' %}
                     {% else %}
--- a/src/content/pages/index.html
+++ b/src/content/pages/index.html
@@ -17,7 +17,7 @@ eleventyExcludeFromCollections: true
         <div class="span-1-7 content-container">
             <div class="">
                 <div class="section-head">
-					{% if site.homepage.animation.enabled %}
+			{% if site.homepage.animation.enabled %}
                     	<h1 class="visually-hidden">{{ site.homepage.title }}</h1>
                         {% include 'components/header-animation.njk' %}
                     {% else %}
--- a/src/content/pages/index.html
+++ b/src/content/pages/index.html
@@ -17,8 +17,8 @@ eleventyExcludeFromCollections: true
         <div class="span-1-7 content-container">
             <div class="">
                 <div class="section-head">
-                    <h1 class="visually-hidden">{{title}}</h1>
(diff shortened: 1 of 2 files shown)


nzakas:
I’m not sure what the intention was here. @SaraSoueidan?

SaraSoueidan:
@nzakas This is the part that was implemented part of the animation.

The intention was to expose a SR-only heading that is _separate_ from the animated heading. So instead of moving the `aria-hidden` as suggested here, the fix should be to expose the heading text in the currently-empty `h1` above the animated heading in the markup.

I believe @cassieevans ’s intention here was to separate them, but for some reason it seems that the SR-only heading is being empty.

Screen Shot 2022-06-28 at 05 08 12

BenDMyers:
Great context - thank you, @SaraSoueidan! I can make that change.

SaraSoueidan:
Thank you @BenDMyers ✨

src/content/pages/index.html
```diff
@@ -17,8 +17,8 @@
         <div class="span-1-7 content-container">
             <div class="">
                 <div class="section-head">
-                    <h1 class="visually-hidden">{{title}}</h1>
-                    {% if site.homepage.animation.enabled %}
+					{% if site.homepage.animation.enabled %}
+                    	<h1 class="visually-hidden">{{ site.homepage.title }}</h1>
```
BenDMyers, line 21: This change moves the visually hidden `h1` inside the `if` check, but that's because in the `else`, there's already an unbroken `h1` that's sufficiently exposed to assistive technologies. Moving this keeps us from unnecessarily rendering extra `h1`s.

src/content/pages/index.html
```diff
@@ -17,8 +17,8 @@
         <div class="span-1-7 content-container">
             <div class="">
                 <div class="section-head">
-                    <h1 class="visually-hidden">{{title}}</h1>
-                    {% if site.homepage.animation.enabled %}
+					{% if site.homepage.animation.enabled %}
```
nzakas, line 20:
```suggestion
			{% if site.homepage.animation.enabled %}
```

src/content/pages/index.html
```diff
@@ -17,8 +17,8 @@
         <div class="span-1-7 content-container">
             <div class="">
                 <div class="section-head">
-                    <h1 class="visually-hidden">{{title}}</h1>
-                    {% if site.homepage.animation.enabled %}
+			{% if site.homepage.animation.enabled %}
```
nzakas, line 20:
```suggestion
		{% if site.homepage.animation.enabled %}
```

nzakas (approved):
LGTM. Thanks.

PR #253, opened 2022-07-04
chore: Enable Nunjucks throwOnUndefined

btmills:
This will throw if we're missing any parameters or accidentally reference substitutions that don't exist. As you'll see in these commits, it catches a number of small accidental omissions.

### DIFF
--- a/src/content/pages/sponsors.html
+++ b/src/content/pages/sponsors.html
@@ -188,8 +188,8 @@ hook: "sponsors_page"
                     <a href="{{ sponsor.url | url }}" class="sponsor-link" rel="nofollow noopener sponsored" target="_blank">
                         <img loading="lazy" decoding="async" src="{{ sponsor.image }}" height="48" width="48"
                             onerror="this.src = '/assets/images/people/_placeholder-avatar.svg'"
-                            alt="{{ sponsor.name }} is donating {{ sponsor.monthlyDonation | dollars }} each month"
-                            title="{{ sponsor.name }} is donating {{ sponsor.monthlyDonation | dollars }} each month" />
+                            alt="{{ sponsor.name if sponsor.name else "Anonymous" }} is donating {{ sponsor.monthlyDonation | dollars }} each month"
+                            title="{{ sponsor.name if sponsor.name else "Anonymous" }} is donating {{ sponsor.monthlyDonation | dollars }} each month" />
                     </a>
                 </li>
                 {% endfor %}
--- a/src/content/pages/sponsors.html
+++ b/src/content/pages/sponsors.html
@@ -14,7 +14,7 @@ hook: "sponsors_page"
 {% set supporting_text = supporting_text | replace("BACKER_COUNT", sponsors.totals.sponsorCount) | replace("MONTHLY_DONATIONS", monthly_donations) %}
 
 {{ hero({
-    title: title,
+    title: site.sponsors_page.title,
     supporting_text: supporting_text,
     buttons: {
         primary: {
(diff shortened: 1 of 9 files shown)


btmills:
I was able to deduce the missing fr string. The es and ja sites are missing the [`from_the_blog`](https://github.com/eslint/eslint.org/blob/220f337003b759c2b95151f9b1c64f53c8142e86/src/_data/sites/en.yml#L285) string. Adding those will allow all sites to pass. @AlixFachin and @luifermoron, would you be able to provide translations for the string I've linked above?

AlixFachin:
> I was able to deduce the missing fr string. The es and ja sites are missing the [`from_the_blog`](https://github.com/eslint/eslint.org/blob/220f337003b759c2b95151f9b1c64f53c8142e86/src/_data/sites/en.yml#L285) string. Adding those will allow all sites to pass. @AlixFachin and @luifermoron, would you be able to provide translations for the string I've linked above?

Sure - will have a look today!
Should I create a brand new PR with the answer?

luifermoron:
> I was able to deduce the missing fr string. The es and ja sites are missing the [`from_the_blog`](https://github.com/eslint/eslint.org/blob/220f337003b759c2b95151f9b1c64f53c8142e86/src/_data/sites/en.yml#L285) string. Adding those will allow all sites to pass. @AlixFachin and @luifermoron, would you be able to provide translations for the string I've linked above?

Sure, I will do it. But now I am facing a COVID situation. I will work on it in following days.

nzakas:
Nice, thanks for doing this @btmills 

@AlixFachin yes please.

@luifermoron take care of yourself.

Worst case scenario is we just use English in the short term.

AlixFachin:
@btmills @nzakas 
Ok I added the translation for the snippet - (最新のブログ記事, which means "recent blog articles")
The PR is located there:
<https://github.com/eslint/eslint.org/pull/259>

I hope this helps!

btmills:
Thank you, @AlixFachin - that worked! I rebased this branch now that #259 is merged, and the ja build is fixed.

The new pt-br site only had one typoed key, and it's working after a quick fix too.

I'll leave this for a few days to give luifermoron some time and add a temporary English fallback if that's not long enough.

nzakas:
“From the blog” only ever appears on blog pages, which never show up on the translation sites. So I think we are good with just using an English placeholder.

btmills:
Thank you, @luifermoron - rebased and everything's green!

Marked as ready for review. This is probably easiest to review one commit at a time, and I've included explanations in most commit messages.

src/content/pages/sponsors.html
```diff
@@ -9,6 +9,7 @@
 {%- from 'components/donation.macro.html' import donationItem %}
 {%- from 'components/hero.macro.html' import hero %}
 
+{% set title = site.sponsors_page.title %}
```
nzakas, line 12: This shouldn’t be necessary? The page title is pulled using the hook value in base.html.

btmills: Disclaimer: my knowledge of Eleventy is fairly limited.

If I remove this line, `title` is undefined.

I see `page_title` being pulled based on `hook` in `base.html`:

https://github.com/eslint/eslint.org/blob/6a26fa2c26d7f332ef2e564da6d753818bda5dca/src/_includes/layouts/base.html#L14

If I remove this line and change `hero({ title: title })` to `hero({ title: page_title })`, `page_title` is also undefined. Should `page_title` be in scope here from `base.html`?

nzakas: Ah I see. In this case, the call to hero should be passing `site.sponsors_page.title` instead of just `title`.

btmills: Gotcha! I updated to do that in ab9854d. I also rebased on `main` to make sure this has the latest changes.

nzakas (approved):
LGTM. Huge win with this.

PR #254, opened 2022-07-04
docs: update installation command

snitin315:
Refers https://github.com/eslint/eslint/issues/16050

We decided to use `npm init @eslint/config`.

Not sure if `@` is clear in this font.

Screenshot 2022-07-04 at 4 30 55 PM

Latest screenshot - 

Screenshot 2022-07-18 at 6 29 53 PM

### DIFF
--- a/src/assets/scss/homepage.scss
+++ b/src/assets/scss/homepage.scss
@@ -46,11 +46,21 @@
         color: var(--headings-color);
     }
 
-    input {
+    .code-snippet {
         all: unset;
         min-width: 30ch;
         color: inherit;
-        font-family: var(--mono-font);      
+        font-family: var(--mono-font);
+        display: inline-block;
+
+        span {
+            font-family: var(--text-font);
+        }
+
+        &[data-copied="true"] {
+            background-color: var(--color-primary-800);
+            color: white;
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


nzakas:
Hmmm yeah. I wonder if we can just Courier for the “@“ symbol?

snitin315:
I tried with [@font-face](https://css-tricks.com/modifying-specific-letters-with-css-and-javascript/#aa-font-face) but wasn't able to make it work 😕 

This is how it looks with `font-family: var(--display-font)`:

Screenshot 2022-07-06 at 9 11 01 AM

nzakas (commented):
This looks nice but I’m worried we are sacrificing accessibility with this approach. I didn’t realize we were using an input field, but it makes sense because keyboard users can tab into the field to read and select text. We may need to think a bit more about this change.

src/assets/scss/homepage.scss
```diff
@@ -46,11 +46,21 @@
...
     }
 
-    input {
+    .code-snippet {
         all: unset;
         min-width: 30ch;
         color: inherit;
-        font-family: var(--mono-font);      
+        font-family: var(--mono-font);
+        display: inline-block;
```
amareshsm, line 54:
```suggestion
```

PR #276, opened 2022-07-25
fix: set reference to view_all_posts

luifermoron:
Just changing the hard-coded word "View all posts" by: view_all_posts for blog page


nzakas (approved):
LGTM. Thanks!

PR #284, opened 2022-07-30
feat: optimize share url button

chenxsan:
Not sure why the share url button is not extended to full width in this case,

1. it's sort of small, I can only see `https://eslint.org/play/#e`,
2. the `vw` is not consistent, it's changing with the size of viewport, and could cause overlapping problem between the texts and copy button, you can see it from the first screenshot

image

Thus I'm proposing some changes to the share url button.

### DIFF
--- a/src/assets/scss/components/playground-configuration.scss
+++ b/src/assets/scss/components/playground-configuration.scss
@@ -152,13 +152,13 @@
 }
 
 .share-url {
-    display: inline-flex;
+    display: flex;
     align-items: center;
     gap: 1rem;
     background-color: var(--lightest-background-color);
     border-radius: var(--border-radius);
     border: 1px solid var(--border-color);
-    padding-right: calc(.625rem + 1vw);
+    padding-right: calc(.625rem + 20px);
 
     max-width: 100%;
     color: var(--headings-color);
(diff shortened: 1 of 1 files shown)


src/assets/scss/components/playground-configuration.scss
```diff
@@ -152,13 +152,13 @@
...
 .share-url {
-    display: inline-flex;
+    display: flex;
     align-items: center;
     gap: 1rem;
     background-color: var(--lightest-background-color);
     border-radius: var(--border-radius);
     border: 1px solid var(--border-color);
-    padding-right: calc(.625rem + 1vw);
+    padding-right: calc(.625rem + 20px);
```
chenxsan, line 161: The share button is 20px width.

kecrily: `px` is an absolute length unit. If it is used, then `calc()` makes no sense.

Ref https://developer.mozilla.org/en-US/docs/Web/CSS/calc#accessibility_concerns

---

Also, can you explain why the `padding` needs to be modified? To solve the problem that share input does not have full width, it seems that you only need to change `inline-flex` to `flex`

kecrily (approved):
Looks good to me

snitin315 (approved):
Nice enhancement. Thank you.

Issue #283, opened 2022-07-28
Update /team page with new team members


nzakas:
We recently created two new teams, and we need to get those displayed on the /team page:

* Website Team - The people who maintain eslint.org.
* Support Team - The people who answer questions in Discord, on our mailing lists, and in discussions.

These should appear, in this order, on the /team page.

## Steps

- [ ] Update [`tools/fetch-team-data.js`](https://github.com/eslint/eslint.org/blob/main/tools/fetch-team-data.js) - I believe you just need to include the new teams in [this section](https://github.com/eslint/eslint.org/blob/main/tools/fetch-team-data.js#L37-L57) and then update [here](https://github.com/eslint/eslint.org/blob/main/tools/fetch-team-data.js#L100-L107). This will generate the `src/_data/team.json` file.
- [ ] Update the [HTML template](https://github.com/eslint/eslint.org/blob/main/src/content/pages/team.html)
- [ ] Add the strings for the team name and descriptions to each file in the `src/_data/sites` directory. It's okay to use English for all of the sites -- we will need to get those translated later.

snitin315:
I'll work on it.

nzakas:
Thanks!

PR #285, opened 2022-07-30
feat: add website and support team on team page

snitin315:
Resolve https://github.com/eslint/eslint.org/issues/283

Screenshot 2022-07-30 at 9 35 40 AM

### DIFF
--- a/.github/workflows/data-fetch.yml
+++ b/.github/workflows/data-fetch.yml
@@ -3,25 +3,15 @@ name: Data Fetch
 on:
   schedule:
   - cron: "0 7 * * *" # Every day at midnight PDT
-  pull_request:
-    types: [labeled]
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
     - name: Check out repo
-      if: github.event_name == 'schedule'
       uses: actions/checkout@v3
       with:
         token: ${{ secrets.WORKFLOW_PUSH_BOT_TOKEN }}
-
-    - name: Check out repo on PR branch
-      if: github.event_name == 'pull_request'
-      uses: actions/checkout@v3
-      with:
-        token: ${{ secrets.WORKFLOW_PUSH_BOT_TOKEN }}
-        ref: ${{ github.event.pull_request.head.ref }}
 
     - name: Set up Node.js
       uses: actions/setup-node@v3
@@ -50,13 +40,6 @@ jobs:
         git config user.email "<eslint@googlegroups.com>"
     
     - name: Save updated files
-      if: github.event_name == 'schedule'
       run: |
         chmod +x ./tools/commit-data.sh
         ./tools/commit-data.sh
-
-    - name: Commit updated data
-      if: github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'fetch data')
-      uses: EndBug/add-and-commit@v9.0.0
-      with:
-        message: "chore: fetch latest data"
--- a/.github/workflows/data-fetch.yml
+++ b/.github/workflows/data-fetch.yml
@@ -4,8 +4,7 @@ on:
   schedule:
   - cron: "0 7 * * *" # Every day at midnight PDT
   pull_request:
-    types:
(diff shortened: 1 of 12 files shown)


.github/workflows/data-fetch.yml
```diff
@@ -3,16 +3,27 @@ name: Data Fetch
 on:
   schedule:
   - cron: "0 7 * * *" # Every day at midnight PDT
+  pull_request:
+    branches:
+      - main
```
snitin315, line 9: I updated the script to run on PR since I didn't have access to the token locally.

We can revert changes in this file if not required. Or maybe push commit only if there are changes in the `tools/` directory.

src/_data/team.json
```diff
@@ -93,7 +93,7 @@
         {
             "username": "platinumazure",
             "name": "Kevin Partington",
-            "title": "ESLint Alumnus",
+            "title": "ESLint Support Team",
```
snitin315, line 96: He seems to be part of both teams.

harish-sethuraman (approved):
LGTM :)

amareshsm (approved):

amareshsm:
LGMT

.github/workflows/data-fetch.yml
```diff
@@ -3,16 +3,26 @@ name: Data Fetch
 on:
   schedule:
   - cron: "0 7 * * *" # Every day at midnight PDT
+  pull_request:
+    types: [labeled]
```
snitin315, line 7: Trigger on `labeled` event only to avoid unnecessary pushing a commit. Ideally this should be triggered only when working with fetching scripts.

src/_data/team.json
nzakas: Oops yes, I’ll remove him from alumni.

src/content/pages/team.html
```diff
@@ -119,6 +119,75 @@ <h2 class="section-title">{{ site.team_page.committers.title }}</h2>
     </div>
 </section>
 
+<section class="section">
```
nzakas, line 122: For each of these sections, I think we should check to make sure there is at least one member before rendering the section.

.github/workflows/data-fetch.yml
nzakas: I’m not sure I understand the intention of this. Can you explain how you envision this working?

nzakas (changes requested):
Thanks for taking this one. Just a couple comments.

.github/workflows/data-fetch.yml
snitin315: I am not able to run the `fetch:team` script locally because I don't have the `ESLINT_GITHUB_TOKEN` available in the local env. So I updated this workflow to leverage it on my PR.

The workflow will be:
- Add `fetch data` label to PR.
- The workflow will be triggered and the latest data will be pushed to the PR branch. See this commit https://github.com/eslint/eslint.org/pull/285/commits/fd9b3159124a4885f54e843a86f6dfb141d9906e for example it wash pushed automatically by the workflow.

snitin315: This is useful when contributors are working on such fetch scripts and don't have access to the token locally.

nzakas: Ah, I see. I’m not sure this is necessary. I’m pretty sure all anyone needs to do is generate a token with org:read permissions. All the data is public so it should be available in the API for everyone. The only reason we use a token is because of the number of requests — I was afraid of hitting the request limit for unauthenticated API calls.

I’m also not sure that this label approach will work for external contributions that are made from forks.

snitin315: We can revert changes to this file.

src/content/pages/team.html
snitin315: @nzakas Done, I added a check - `{% if team.teamName.length > 0 %}`

nzakas (approved):
LGTM. Thanks!

PR #286, opened 2022-07-30
build: enable webpack persistent cache

snitin315:
Related docs - https://webpack.js.org/configuration/cache/

With cache, the webpack re-build time is now reduced from `12349 ms` to `323 ms`. 

Initial build without cache:
Screenshot 2022-07-30 at 9 51 56 AM

Re-build with cache
Screenshot 2022-07-30 at 9 52 11 AM

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -11,6 +11,13 @@ module.exports = {
         filename: "[name].js",
         path: path.resolve(__dirname, "_site/assets/js")
     },
+    cache: {
+        type: "filesystem",
+        buildDependencies: {
+            config: [__filename]
+        },
+        cacheDirectory: path.resolve(__dirname, "node_modules/.cache/webpack")
+    },
     resolve: {
         extensions: [".js", ".jsx"],
         mainFields: ["browser", "main", "module"]
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -15,8 +15,7 @@ module.exports = {
         type: "filesystem",
         buildDependencies: {
             config: [__filename]
-        },
-        cacheDirectory: path.resolve(__dirname, "node_modules/.cache/webpack")
+        }
     },
     resolve: {
         extensions: [".js", ".jsx"],


webpack.config.js
```diff
@@ -11,6 +11,13 @@ module.exports = {
         filename: "[name].js",
         path: path.resolve(__dirname, "_site/assets/js")
     },
+    cache: {
+        type: "filesystem",
+        buildDependencies: {
+            config: [__filename]
+        },
+        cacheDirectory: path.resolve(__dirname, "node_modules/.cache/webpack")
```
kecrily, line 19: Does this seem to be consistent with the default value, and is there a need for explicit configuration?

Ref https://webpack.js.org/configuration/cache/#cachecachelocation

snitin315: Yes this is the default value. We can omit this if we want.

webpack.config.js
```diff
@@ -11,6 +11,13 @@ module.exports = {
         filename: "[name].js",
         path: path.resolve(__dirname, "_site/assets/js")
     },
+    cache: {
+        type: "filesystem",
```
harish-sethuraman, line 15: https://webpack.js.org/configuration/cache/#cachecompression

Shall we also try to add `brotli` compression since we have the type for cache as `file system`

snitin315: In development mode?

snitin315: `production` mode uses `gzip` compression by default. I believe webpack defaults are good.

harish-sethuraman: In general we can use the combination of gzip + brotli to serve files efficiently. Brotli saves us more space. 

> The advantage for Brotli over gzip is that it makes use of a dictionary and thus it only needs to send keys instead of full keywords. According to [certsimple](https://certsimple.com/blog/nginx-brotli),
> 
> Javascript files compressed with Brotli are 14% smaller than gzip.
> HTML files are 21% smaller than gzip.
> CSS files are 17% smaller than gzip.

https://tech.oyorooms.com/how-brotli-compression-gave-us-37-latency-improvement-14d41e50fee4

nzakas: I think the goal here is to help local development rather than deploy compressed assets, so I think the Webpack defaults are fine.

webpack.config.js
nzakas: If it’s the same value, I’m in favor of omitting.

amareshsm (approved):

amareshsm:
LGTM

webpack.config.js
snitin315: Updated.

nzakas (approved):
LGTM

PR #292, opened 2022-08-02
build: Add pre commit hook

amareshsm:
**Description:** 
1. Added `pre-commit` hook with `yorkie` to lint staged files and minify SVG files.
2. Added scripts to lint `.{js,jsx}` and `.scss` files
3. Added script to minify SVG files using `svgo`.


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx lint-staged
```
nzakas, line 4: This won’t work on non-Linux systems.

nzakas:
Thanks for taking a stab at this. A few things:

1. please provide a complete description when submitting a pull request. The description should list out all the changes you made so that it’s easy to understand what you did. Can you please update the description?
2. Changes to build system should be either “chore” or “build”. “feat” is only for end-user facing changes. 
3. We generally use Yorkie instead of Husky for precommit hooks.

package.json
```diff
@@ -28,7 +28,30 @@
         "start": "npm-run-all build:sass --parallel watch:*",
         "build": "npm-run-all install:playground build:sass build:eleventy build:webpack images",
         "lint": "eslint --ext=.js,.jsx .",
-        "fix": "npm run lint -- --fix"
+        "fix": "npm run lint -- --fix",
+        "lint:scss": "stylelint '**/*.scss'",
+        "lint:scss:fix": "stylelint '**/*.scss' --fix",
```
harish-sethuraman, line 33: Please try to re use the lint command with additional arg like `lint`. ie: `npm run lint:scss -- --fix` so that next time if we change the directories with scss to lint we don't want to change the fix command too?

package.json
```diff
@@ -28,7 +28,30 @@
...
+        "prepare": "husky install",
+        "svgo": "svgo -r -f ./"
```
harish-sethuraman, line 35: Should we rename this command to be something meaningful? We have named the command for minifying images as `images`.

package.json
```diff
@@ -28,7 +28,30 @@
...
+    },
+    "lint-staged": {
+        "**/*.{js,jsx}": [
+            "eslint --fix",
+            "git add"
```
harish-sethuraman, line 40: Do we need a `git add` here? Ideally we should allow users to git add necessary files ?

package.json
amareshsm: yes, we can use a meaningful name.

package.json
amareshsm: done

package.json
amareshsm: I tough we can add those files after fixing them. Better we can allow users to add necessary files.

amareshsm:
Fixing the lint error. I will update this PR soon.

package.json
```diff
@@ -27,8 +27,21 @@
...
         "start": "npm-run-all build:sass --parallel watch:*",
         "build": "npm-run-all install:playground build:sass build:eleventy build:webpack images",
-        "lint": "eslint --ext=.js,.jsx .",
-        "fix": "npm run lint -- --fix"
+        "minify:svg": "svgo -r -f ./",
+        "lint:js": "eslint --ext=.js,.jsx .",
+        "lint:js:fix": "npm run lint:js -- --fix",
+        "lint:scss": "stylelint '**/*.scss'",
+        "lint:scss:fix": "npm run lint:scss --fix",
+        "lint": "npm-run-all --parallel lint:js lint:scss",
```
harish-sethuraman, line 35:
```suggestion
        "lint": "npm-run-all --parallel lint:*",
```

package.json
```diff
@@ -27,8 +27,21 @@
...
+        "fix": "npm-run-all --parallel lint:js:fix lint:scss:fix"
```
harish-sethuraman, line 36:
```suggestion
        "fix": "npm-run-all --parallel lint:*:*"
```

I think we can do it this way? Can you once verify if this works?

package.json
```diff
@@ -28,8 +28,21 @@
         "fetch": "npm-run-all --parallel fetch:*",
         "start": "npm-run-all build:sass --parallel watch:*",
         "build": "npm-run-all install:playground build:sass build:eleventy build:webpack images",
-        "lint": "eslint --ext=.js,.jsx . && markdownlint \"**/*.md\" -i node_modules",
-        "fix": "npm run lint -- --fix"
+        "minify:svg": "svgo -r -f ./",
```
harish-sethuraman, line 31: wondering how this would work? We don't actually have the svgo dep installed in the repo and we are running it via npx?

harish-sethuraman: Note: seems like svgo comes as deps via imagemin package so there is no need to install it and works out of the box

harish-sethuraman (approved):

package.json
amareshsm: yeah

nzakas (approved):
LGTM. Thanks!

PR #294, opened 2022-08-03
fix: invoke `focus` only on recently added rule

shanpriyan:
- Removed `.focus()` and `.scrollIntoView()` call on initial mount of every `RuleItem` in the list
- Focus new rule and scrollIntoview only after adding a new rule


snitin315:
> Invoked focus only on the last RuleItem in the list, instead of calling on every item

Why focus only on the last item? We sort rule items alphabetically so added rule will not always be the last item on the list.

shanpriyan:
> > Invoked focus only on the last RuleItem in the list, instead of calling on every item
> 
> Why focus only on the last item? We sort rule items alphabetically so added rule will not always be the last item on the list.

Sorry, I have considered the initial mount scenario of rule list where it focuses the last rule item. It will be a problem after adding a new rule to list. Will commit new changes.

Should we focus the last rule in the list after initial mount of the playground? Is that the expected behaviour ?

snitin315:
What is the issue with the current behavior? Currently it focuses the rule which was added and scrolls till the rule is visible which is the expected behavior.

shanpriyan:
> What is the issue with the current behavior? Currently it focuses the rule which was added and scrolls till the rule is visible which is the expected behavior.

There is no issue in the current behaviour after adding a new rule, I was asking about **after the initial mount** of the page where it scrolls to the last item and focuses it, Is that an expected one?

amareshsm:
Before changes:
When we open the playground last rule input is focused and scrolled into view automatically.

After changes:
 Scroll into view and Focus the rule input element whenever a new rule is added.
 
 @shanpriyan Correct me if I'm wrong.

shanpriyan:
> Before changes: When we open the playground last rule input is focused and scrolled into view automatically.
> 
> After changes: Scroll into view and Focus the rule input element whenever a new rule is added.
> 
> @shanpriyan Correct me if I'm wrong.

Yes, correct

src/playground/components/RuleItem.js
```diff
@@ -1,16 +1,11 @@
...
-        <li ref={ref} className="config__added-rules__item">
+const RuleItem = ({ children }) =>
+    (
+        <li className="config__added-rules__item">
             {children}
         </li>
     );
-}
+
+RuleItem.displayName = "RuleItem";
```
harish-sethuraman, line 10:
```suggestion
```
Do we need this? We are already exporting an method devtools should pick its name?

shanpriyan: This was needed when `forwardRef` was used. Removed it now

src/playground/components/RuleItem.js
```diff
@@ -1,16 +1,5 @@
...
-        ref.current.scrollIntoView({ behavior: "smooth", block: "end" });
-        ref.current.children[1].focus();
-    }, []);
-
-    return (
-        <li ref={ref} className="config__added-rules__item">
-            {children}
-        </li>
-    );
+    return <li className="config__added-rules__item">{children}</li>;
```
harish-sethuraman, line 4: Just wondering if we'd need a separate file for this one line? This could be inside configuration itself? We don't want this to be a component? WDYT?

shanpriyan: Yes, we can move it to `Configuration`. It doesn't make sense to keep it in a separate file.

harish-sethuraman (approved):
LGTM

amareshsm (approved):

amareshsm:
LGTM

src/playground/components/Configuration.js
```diff
@@ -219,14 +219,13 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
                                             delete options.rules[ruleName];
                                             setRulesWithInvalidConfigs(new Set([...rulesWithInvalidConfigs].filter(rule => rule !== ruleName)));
                                             onUpdate(Object.assign({}, options));
-                                        }}
+                                        } }
```
snitin315, line 222:
```suggestion
                                        }}
```

src/playground/components/Configuration.js
```diff
@@ -219,14 +219,13 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
...
                                     >
                                         <svg width="25" height="25" viewBox="0 0 45 44" fill="none" xmlns="http://www.w3.org/2000/svg">
-                                            <path d="M28.5 16L16.5 28M16.5 16L28.5 28" stroke="var(--link-color)" strokeWidth="5" strokeLinecap="round" strokeLinejoin="round"/>
+                                            <path d="M28.5 16L16.5 28M16.5 16L28.5 28" stroke="var(--link-color)" strokeWidth="5" strokeLinecap="round" strokeLinejoin="round" />
```
snitin315, line 225:
```suggestion
                                            <path d="M28.5 16L16.5 28M16.5 16L28.5 28" stroke="var(--link-color)" strokeWidth="5" strokeLinecap="round" strokeLinejoin="round"/>
```

src/playground/components/Configuration.js
```diff
@@ -219,14 +219,13 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
...
                                         </svg>
                                     </button>
-                                </h4>
-                                <input
+                                </h4><input
```
snitin315, line 228:
```suggestion
                                </h4>
                                <input
```

src/playground/components/Configuration.js
```diff
@@ -240,14 +239,13 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
                                         } catch {
                                             setRulesWithInvalidConfigs(new Set([...rulesWithInvalidConfigs, ruleName]));
                                         }
-                                    }}
-                                />
+                                    } } />
```
snitin315, line 242: This can be reverted too.

snitin315 (commented):
I see some unnecessary formatting changes, which should be reverted. Otherwise, LGTM

snitin315 (approved):
LGTM, Thanks for the PR 💯

Issue #183, opened 2022-03-26
Bug: Flags don't show up in language selector in Chromium


nzakas:
### URL(s)

https://es.eslint.org
https://es.eslint.org/languages

### What did you do?

Scroll all the way to the bottom to the language picker.

### What did you expect to happen?

To see a flag next to "Spanish (ES)" as in Firefox:
![firefox-language-picker](https://user-images.githubusercontent.com/38546/160222623-3ebf2e66-4ec2-4d22-a382-320f7375421b.png)

### What actually happened?

I just saw "ES" instead of the flag. From Chrome:
![chrome-language-picker](https://user-images.githubusercontent.com/38546/160222635-2214ce51-7ff2-4440-b438-307791d4cedf.png)

### Participation

- [X] I am willing to submit a pull request for this issue.

### Additional comments

This appears on all language sites, not just Spanish.

nzakas:
@SaraSoueidan do you have any idea what might be happening here?

SaraSoueidan:
No, that’s odd. I just checked on my macOS and in my Windows VM and it seems that this is happening only on Chrome on Windows. I'm not sure why. I've never come across this before. 🤔

SaraSoueidan:
Apparently it's a known Windows 10 issue: https://answers.microsoft.com/en-us/windows/forum/all/flag-emoji/85b163bc-786a-4918-9042-763ccf4b6c05

Do you get the same problem in Windows 11? I can't test on Win 11 because I don't have its VM yet.

nzakas:
I haven’t upgraded to Windows 11, so I’ll not sure. It seems like there’s not a good solution here. Maybe worth filing a bug on Chrome.

SaraSoueidan:
I think others have already done that. It seems to be a Windows issue more so than a Chrome issue.

nzakas:
It works in Firefox on Windows, so it must be solvable.

PR #302, opened 2022-08-05
fix: `Uncaught TypeError` trying to access body from head tag in windows

### LINKED ISSUE #301: Bug:  Uncaught TypeError: Cannot read properties of null (reading 'classList') in windows
### URL(s)

https://eslint.org/play/

### What did you do?

1. On `windows`  OS
1. Open any browser and visit [playground](https://eslint.org/play/)
2. Open `console` in devtools

- No errors in console
- Should add `win` to the body class, if site is opened on `windows` machine

### What actually happened?

Received `document.body` is null error in console

### Participation

- [X] I am willing to submit a pull request for this issue.

## Console
image
## Source
image

### Cause
Trying to access `document.body` from `head` tag https://github.com/eslint/eslint.org/blob/3fdd3964c4383b1f970f550aa887795a01bcc2d5/src/_includes/layouts/base.html#L128

shanpriyan:
Moved the windows `userAgent` check to body tag

Resolves #301


src/_includes/layouts/base.html
```diff
@@ -163,6 +160,11 @@
     <script src="{{ '/assets/js/tabs.js' | url }}"></script>
     <script src="{{ '/assets/js/components-index.js' | url }}"></script>
     {%- endif -%}
+    <script>
```
harish-sethuraman, line 163: Just wondering why we'd need this piece of code? Are we applying any special style or using it somewhere? else this is totally un necessary?

shanpriyan: Seems like there is no special styling exists as of now and not able to find where this is being used also.

amareshsm: I guess we don't need this now. Previously this logic was added to set specific `font-family` for windows. After some time that styling code was removed.

shanpriyan: Shall I remove the userAgent check, then?

amareshsm: @kecrily Can you confirm this?

amareshsm:
https://github.com/eslint/eslint.org/issues/301

src/_includes/layouts/base.html
nzakas: This is needed to fix an issue with emoji flags not showing up in Chromium on Windows:
https://github.com/eslint/eslint.org/issues/183

snitin315 (approved):
LGTM

src/_includes/layouts/base.html
harish-sethuraman: I don’t see that #213 makes use of the class win to show emojis? Or am I missing something 😅

kecrily: Yes. It was used to restrict twemoji to Windows and use the default emoji for other platforms. I was later told that twemoji should be used for all platforms. This code seems useless now, let's remove it.

kecrily: > I don’t see that #213 makes use of the class win to show emojis? Or am I missing something 😅

It's in #215

src/_includes/layouts/base.html
```diff
@@ -163,6 +160,11 @@
...
+        if (/windows/i.test(navigator.userAgent)) {
+            document.body.classList.add('win')
+        }
+    </script>
```
kecrily, line 167:
```suggestion
```

src/_includes/layouts/base.html
snitin315: Yes, if it's not used let's remove 👍🏻

harish-sethuraman:
So previously we were using the class `.win` for applying emoji flags fonts for windows alone in this PR #215 but then this was reverted due to other issues but the win classlist addition code alone came back after the revert which is now not needed. So we can now consider removing the code as suggested by others as well :)

harish-sethuraman (approved):
LGTM Thanks ✨

amareshsm:
In https://github.com/eslint/eslint.org/issues/183 PR emoji flag issue was fixed and a script was added to add the `win` class name for windows. Styles for that class were not added in the https://github.com/eslint/eslint.org/issues/183 PR.

later in this https://github.com/eslint/eslint.org/pull/215/ PR script block was moved inside the body and added styles. But that particular flags font family (`"Twemoji Country Flags"`) was only added to the `win` class name. It should include other existing text font families also. 

<table>
<tr>
<td> Styles added in  #215   </td> <td> Expected </td>
</tr>
<tr>
<td>

```css
.win {
    font-family: "Twemoji Country Flags" !important;
}
```

 </td>
<td>

```css
.win {
    font-family: "Inter",
            -apple-system,
            BlinkMacSystemFont,
            "Segoe UI",
            Roboto,
            Helvetica,
            Arial,
            sans-serif,
            "Apple Color Emoji",
            "Twemoji Country Flags",
            "Segoe UI Emoji",
            "Segoe UI Symbol";
}
```

</td>
</tr>
</table>

Since the other text font families were not added incorrect font was displayed in the language picker. So #215 was reverted by #222.

Just added this comment for context. If want to use the system's own flag on non-Windows systems and custom flags for windows systems then we need to add the script and styles. Otherwise, we can remove the script block which is currently not in use.

kecrily (approved):
LGTM

amareshsm (approved):

amareshsm:
LGTM

nzakas (approved):

nzakas:
Thanks for digging into that, everyone. For future reference, when there are multiple people agreeing, please feel free to merge the PR. You don’t need to wait for anyone on the TSC.

PR #303, opened 2022-08-06
feat: update wording of flat config blog post

filiptammergard:
In two places, the config array is referred to as being "returned", while it is rather being exported. While this a minor nitpick, I was confused for a second trying to find what's being returned. An exported array could probably be said to be returned, but since `export` and `return` are JavaScript concepts that are different in many ways, I think it makes sense to use the accurate terms.


nzakas (approved):
Much clearer, thanks.

PR #304, opened 2022-08-08
chore: Add Markdownlint formatting for files

nschonni:
Added initial config and install in first commit, then cleaned up single rules in subsequent ones.


nzakas:
Thanks for looking at this. A bit of complexity: the release notes posts are autogenerated and based on that, they may end up with extra whitespace. I don't want the build to break because of that (it's a completely automatic generate and commit step), which this PR would do as-is. What would you suggest for this?

Also, I believe markdown lint can autofix, could update the `fix` script to include that?

nschonni:
> Thanks for looking at this. A bit of complexity: the release notes posts are autogenerated and based on that, they may end up with extra whitespace. I don't want the build to break because of that (it's a completely automatic generate and commit step), which this PR would do as-is. What would you suggest for this?

If the double blank lines or some of the trailing whitespace rules are problematic for the autogeneration, they could be disabled again. I just fixed up the ones that seem non-controversial 😄 



> Also, I believe markdown lint can autofix, could update the `fix` script to include that?

Yes, the current `npm run fix` will autofix any ones that it can, because it's already passing `--fix` to the `npm run lint` command. I just double checked that it worked, but I didn't confirm that it wasn't stepping on the ESLint command also getting the `--fix` parameter

package.json
```diff
@@ -27,7 +27,7 @@
         "fetch": "npm-run-all --parallel fetch:*",
         "start": "npm-run-all build:sass --parallel watch:*",
         "build": "npm-run-all install:playground build:sass build:eleventy build:webpack images",
-        "lint": "eslint --ext=.js,.jsx .",
+        "lint": "eslint --ext=.js,.jsx . && markdownlint \"**/*.md\" -i node_modules",
```
harish-sethuraman, line 30: I think we can introduce a separate command for linting markdown? something like `lint:md` and then use use `npm-run-all` in parallel to lint js and md files?

kecrily:
I noticed that markdownlint has an ESLint plugin. Maybe we can use it?

nzakas:
@nschonni Cool. Yeah, I think the trailing whitespace and double blank lines rules would cause issues with the autogeneration, so if we can disable those, I think that would be the best choice.

@kecrily I think I'd prefer to stick with the CLI for this.

nschonni:
> @nschonni Cool. Yeah, I think the trailing whitespace and double blank lines rules would cause issues with the autogeneration, so if we can disable those, I think that would be the best choice.

I've re-disabled the MD009 and MD012 rules, but left the original cleanup commits

nzakas (approved):
LGTM. Thanks!

PR #307, opened 2022-08-12
chore: remove old playground's package.json and use main site's one

harish-sethuraman:
~I tried to run the repo and it didnt start since I didn't notice we had two package.json. It would be better to have one package.json ideally.~

preview:
Screenshot 2022-09-07 at 5 47 50 PM

Adding a fallback message that shows up before react app mounts to inform user to run `install:playground` command


snitin315 (changes requested):
This is intentional. We have two package.json so that the playground is built and deployed with the latest version of ESLlint available. 

See [developer setup](https://github.com/eslint/eslint.org#developer-setup), it mentions running `npm run install:playground`.

harish-sethuraman:
How about we use one common ESLint in the whole repo and it would be latest? Just wondering why we’d need two package.json

snitin315:
Using "latest" is problematic because we will end up getting lint errors on CI if a release introduces a breaking change.

snitin315:
One solution is to add a `postinstall` script to run `npm run install:playground`, see [related discussion here](https://github.com/eslint/eslint.org/pull/268#discussion_r923970396).

harish-sethuraman:
Added postinstall and it works fine

nzakas:
We had already discussed and I’m not a fan of postinstall for this purpose. There’s no reason to go through the long playground build process unless you want to work on the playground.

snitin315:
> We may also want to add an error message in playground.html that explains to run install:playground ( https://github.com/eslint/eslint.org/pull/268#discussion_r925066449 )

I feel this is what we need to add here.

harish-sethuraman:
> There’s no reason to go through the long playground build process unless you want to work on the playground.

Just wondering? We are just installing two deps in playground and aren't running any build? 

> We may also want to add an error message in playground.html that explains to run install:playground ( https://github.com/eslint/eslint.org/pull/268#discussion_r925066449 )
>I feel this is what we need to add here.

Should we detect if deps (eslint) is present or not and then remove the error message shown?

nzakas:
I don’t think we need to do any detection. Just include a static message in the playground.html file that will be overwritten if the React app loads.

harish-sethuraman (commented):
Changed to show prompt to install dependencies...

amareshsm (approved):
LGTM

nzakas (approved):
Nice!

nzakas:
@snitin315 can you please review this so we can merge?

snitin315 (approved):
LGTM

Issue #278, opened 2022-06-25
Suggestion: Improve UX of rules managing in Playground


pubmikeb:
### What problem do you want to solve?

Since the https://eslint.org redesign, in order to add a rule in Playground you need to add it manually:
1. Click on `Choose a rule` combo box
2. Find a desired rule in a very long list of all available rules
3. Click on it
4. Click on `Add this rule`

This four steps you need to repeat manually for every rule you want to add. That's a tedious and time burning work.

### What do you think is the correct solution?

Just use an UX from the previous version of https://eslint.org, where you could just click on "Enable all rules" and moreover, all rules have been grouped by the types with a good visualization, which made it extremely easy to maintain an ESLint config.

Poor UX. It would take me hours to find and add all required rules:
![25_121209](https://user-images.githubusercontent.com/17231333/175769138-473e2fd2-56ab-4562-9e95-5e47329d05d1.png)
Plus, everytime I click on `Add this rule`, page is automatically scrolled to the place, where a rule has been added, then I need manually scroll back to the head of a page to add another rule. It's a torture!

Good UX. Just with several clicks get a perfect `eslintrc.json`:
![25_120844](https://user-images.githubusercontent.com/17231333/175769068-c4d7977a-efa0-4af7-bef8-6d9ef30085ff.png)
And the UI is clean, spacious and not to overloaded, in contrast to the new one.

### Participation

- [ ] I am willing to submit a pull request for this change.

### Additional comments

Please, consider returning the former Playground UX, so we could generate `eslintrc.json` within several clicks.

harish-sethuraman:
Are we considering to use a multi select for dropdown to add rules? If yes I can pick this up.

nzakas:
@harish-sethuraman I think that depends on the design. As already mentioned, we don’t want to go back to checkboxes. If you have an idea for what it might look like, I’d suggest posting a design first before attempting to implement anything.

harish-sethuraman:
<img width="563" alt="Screenshot 2022-08-09 at 10 02 01 AM" src="https://user-images.githubusercontent.com/32865581/183564776-aee6b7fe-e57e-4872-9832-a5d7092d6b4c.png">
@nzakas We are not going to move to checkboxes. We will use select that allows us to select multiple options like shown above.

nzakas:
@harish-sethuraman I’m not sure how that solves the problem. Can you explain more about what the user experience would be like?

harish-sethuraman:
[...]

PR #308, opened 2022-08-12
feat: add multiselect to choose rules

harish-sethuraman:
Fixes https://github.com/eslint/eslint.org/issues/278


nzakas:
Ah, I see what you mean. Some thoughts:

1. "Add this rule" should probably change to "Add these rules" if we can add a bunch at once.
2. It also seems like we should have a keyboard way to add all the selected rules. Maybe Ctrl+Enter could add all of the rules that were selected?
3. I'm concerned that adding a bunch of rules at once means they will be difficult to find in the list. When we added one rule at a time, we always set focus to the rule that was added.
4. I wonder if this actually solves the problem mentioned in #278? If the concern is taking too long to add multiple rules, are we sure this is better?

harish-sethuraman:
> "Add this rule" should probably change to "Add these rules" if we can add a bunch at once.
 - It shows "Add this rule" when one or less than one rule is selected. Else it shows "Add these rules" this is the behaviour right now with my change.
> It also seems like we should have a keyboard way to add all the selected rules. Maybe Ctrl+Enter could add all of the rules that were selected?
- Just like right now we can move the focus to "Add these rules" and hit enter to add all selected rules. Or we can add Ctrl + Enter inside the select itself to do this. 

> I'm concerned that adding a bunch of rules at once means they will be difficult to find in the list. When we added one rule at a time, we always set focus to the rule that was added.
- I agree on this we would be able to highlight the first rule and focus that alone :( 

> I wonder if this actually solves the problem mentioned in https://github.com/eslint/eslint.org/issues/278? If the concern is taking too long to add multiple rules, are we sure this is better?
- Yes, https://github.com/eslint/eslint.org/issues/278#issuecomment-1194849929

nzakas:
I'd still like to see Ctrl + Enter. :) I prefer not making people tab around if it's not necessary.

harish-sethuraman:
Im not sure if `react-select` library allows such key strokes for doing some actions while its input field is active... We might want to add a global keystroke listener for listening to `Ctrl + Enter`? or is there any other way we could go with in case anyone knows 🤔

nzakas:
I'm not familiar with react-select, but there must be a way to know when a key is pressed?

nzakas:
@harish-sethuraman are you still working on this?

harish-sethuraman:
Hey sorry totally missed this. 

I think we can replace the input field with our own field so that we can have control over keypress. I'll look into it this week.

nzakas:
Thanks!

harish-sethuraman:
This is sorted and ready for review

src/playground/components/Configuration.js
```diff
@@ -105,10 +105,35 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
...
+        if (props.isHidden) {
+            return <components.Input {...props} />;
+        }
+        return (
+            <components.Input {...props} onKeyDown={e => {
+                if ((e.metaKey || e.ctrlKey) && e.code === "Enter") {
+                    handleRuleChange();
+                } else {
+                    props.onKeyDown(e);
+                }
```
snitin315, line 132: When we press `cmd + enter` when the dropdown is present it doesn't add all the rules but selects one rule from the dropdown and scrolls to one of the selected rules.

https://user-images.githubusercontent.com/46647141/204116153-999f2499-4d39-4dc9-a09d-f052c83910d9.mov

harish-sethuraman: weird... will check

src/playground/components/Configuration.js
```diff
@@ -207,31 +232,29 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
...
                             } else {
-                                setSelectedRule(selected.value);
+                                setSelectedRules(selected.map(values => values.value));
                             }
                         }}
                         options={ruleNamesOptions}
                     />
                     <button
                         className="c-btn c-btn--ghost add-rule-btn"
                         onClick={() => {
```
shanpriyan, line 253:
```diff
- onClick={() => {
- handleRuleChange();
- }}
+ onClick={handleRuleChange}

File: src/playground/components/Configuration.js (hunk 1 above)
Comment (amareshsm) [reply src/playground/components/Configuration.js:132]: Same in windows also not working.

**Comment by harish-sethuraman** [comment]
I see that it is not possible to add `Command + Enter` for adding rules. There are events for input field how ever when we handle the `onkeydown` the options are added to the selected values how ever the `onChange` is executed after `onKeyDown`. This way we can't trigger the `Command + Enter` change before the option is being selected.

**Comment by nzakas** [comment]
Okay, I still feel like there must be a way to make Ctrl+Enter work, but I don't see any reason to hold up this PR just for that.

Do you think this is in a good enough state to merge now?

**Comment by harish-sethuraman** [comment]
I'd have to revert the changes made for adding ctrl + enter this will be ready tomorrow 👍🏻

**Review by amareshsm — APPROVED** [review]
LGTM. Leaving this open for someone else to double-check

**Comment by siva-kannan3** [comment]
> I see that it is not possible to add `Command + Enter` for adding rules. There are events for input field how ever when we handle the `onkeydown` the options are added to the selected values how ever the `onChange` is executed after `onKeyDown`. This way we can't trigger the `Command + Enter` change before the option is being selected.

@harish-sethuraman, it seems the keydown event you attached to the Input component is one of the inner components of React select. So when we click ```Cmd + Enter```, keydown event of input is called and then event is propagated to react select (Select component also listens for keydown). So if we stop the input event from propagation on ```Cmd + Enter```, the handler you have given is working.

**Comment by harish-sethuraman** [comment]
@siva-kannan3 I remember trying this out.  If it works in all cases then we can add it in this PR or a separate one. Can you share snippet?

**Comment by siva-kannan3** [comment]
> @siva-kannan3 I remember trying this out. If it works in all cases then we can add it in this PR or a separate one. Can you share snippet?

https://codesandbox.io/s/react-select-multi-select-example-rdg1me?file=/example.tsx
@harish-sethuraman

File: src/playground/components/Configuration.js (hunk 1 above)
Comment (harish-sethuraman) [inline src/playground/components/Configuration.js:132]:
```suggestion
                    handleRuleChange();
                    e.stopPropagation();
```

harish-sethuraman:
Will wait for @siva-kannan3 to add suggestion. If it works we can add that in the same PR as well 👍🏻

src/playground/components/Configuration.js
```diff
@@ -109,6 +109,31 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
...
+    const Input = props => {
+        if (props.isHidden) {
+            return <components.Input {...props} />;
+        }
+        return (
+            <components.Input {...props} onKeyDown={e => {
+                if ((e.metaKey || e.ctrlKey) && e.code === "Enter") {
+                    handleRuleChange();
+                } else {
+                    props.onKeyDown(e);
```
siva-kannan3, line 131: this throwing error.
I think we cannot trigger a event like this. 
image

src/playground/components/Configuration.js
```diff
@@ -109,6 +109,31 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
...
+        selectedRules.forEach(selectedRule => {
+            if (ruleNames.includes(selectedRule)) {
+                options.rules[selectedRule] = ["error"];
+                onUpdate(Object.assign({}, options));
+                ruleInputRef.current.setValue("");
+            }
+        });
+    };
+
+    const Input = props => {
```
siva-kannan3, line 122: whenever ```<Configuration/>``` state changes, this ```<Input/>``` component is reinitiated ( since ```<Input/>``` is defined inside ```<Configuration/>```), so every time we choose a rule, we lose focus on the input. So need to handle this as well.

src/playground/components/Configuration.js
```diff
@@ -208,6 +233,7 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
                     <label htmlFor="rules" className="combo-label"><span className="label__text">Choose a rule</span></label>
                     <Select
                         isMulti
+                        components={{ Input }}
```
siva-kannan3, line 236: ```closeOnMultiSelect={false}```
This could be added to all multi select components. since everytime a we choose a rule, the dropdown closes due to closeOnMultiSelect is set to true as default

src/playground/components/Configuration.js
harish-sethuraman: Yea I'll remove that anyways its not needed

src/playground/components/Configuration.js
harish-sethuraman: We have to confirm if this is expected or not. I think this is expected?

src/playground/components/Configuration.js
harish-sethuraman: ✅  will do

shanpriyan: we can keep the `Input` component definition outside the `Configuration` to fix the focus issue

src/playground/components/Configuration.js
```diff
@@ -105,10 +105,35 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
...
+        });
+    };
+
+    const Input = props => {
+        if (props.isHidden) {
+            return <components.Input {...props} />;
+        }
+        return (
+            <components.Input {...props} onKeyDown={e => {
+                if ((e.metaKey || e.ctrlKey) && e.code === "Enter") {
```
shanpriyan, line 128: when `Enter` from numpad is pressed, `e.code` will be `"NumpadEnter"`
```suggestion
                if ((e.metaKey || e.ctrlKey) && e.key === "Enter") {
```

src/playground/components/Configuration.js
harish-sethuraman: I don't think its happening because of this. I tried moving it out of the Configuration component. Seems like the input field is blurred once the option is selected when we use custom input. 😞

shanpriyan: Ah, My bad. Even I tried the same just now it didn't solve the issue. It seems like we need to find a different approach to fix it. 😕

snitin315 (approved):
LGTM, thanks ⭐ .

harish-sethuraman:
Previously you would be able to add one rule at a time now you can select multiple rules and add them at once. But why would you click on select all and add all rules? can you raise this as an issue? we can discuss it there?

PR #314, opened 2022-08-13
fix: build failure

harish-sethuraman:
Seems like bot added null as URL and the build failed :( 

We might want to update it with a valid URL or go with adding URL only when the URL is available.

I've added valid URL for now from the chrome developer blogs since this made sense to me thoughts?

### DIFF
--- a/src/_data/sponsors.json
+++ b/src/_data/sponsors.json
@@ -7,7 +7,7 @@
     "platinum": [
         {
             "name": "Frameworks Fund",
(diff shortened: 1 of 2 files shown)


snitin315:
> adding URL only when the URL is available.

I'm 👍🏻 on this, it will prevent future build failures for such cases.

harish-sethuraman:
Removed URL and added valid checks

btmills (approved):
LGTM, thanks for fixing this already! Merging, then will rebase #315.

PR #321, opened 2022-08-18
feat: add keyboard controls to blog search

### LINKED ISSUE #309: Add keyboard controls to blog search
Right now the blog search is mouse-only. We should update it to include keyboard controls:

1. After typing a couple characters, pressing down should move focus to the first search result and highlight it. You should be able to move to previous and next search results using up/down keys.
2. Pressing ESC should hide the dropdown and set focus back to the textbox.

shanpriyan:
- Added keyboard navigation of blog search results using `ArrowUp` and `ArrowDown`
- Pressing `Enter` on selected result will open the selected page
- Pressing `Esc` key will close the search results dropdown and focuses the input
- Added `⌘ + K` key binding to focus the search field

Closes #309

### DIFF
--- a/src/assets/js/search.js
+++ b/src/assets/js/search.js
@@ -144,12 +144,6 @@ searchInput.addEventListener('keyup', function (e) {
     searchQuery = query
 });
 
-resultsElement.addEventListener('keydown', function(e) {
-    if(e.key === "Escape") {
-        clearSearchResults();
-    }
-}, true);
-
 searchClearBtn.addEventListener('click', function(e) {
     searchInput.value = '';
     searchInput.focus();
@@ -170,6 +164,10 @@ searchInput.addEventListener('keydown', function (e) {
             break;
         case "Enter":
             resultsElement.querySelector('li.selected a').click();
+            break;
+        case "Escape":
+            e.preventDefault();
+            clearSearchResults();
             break;
 
         default:
--- a/src/assets/js/search.js
+++ b/src/assets/js/search.js
@@ -22,6 +22,9 @@ const resultsElement = document.querySelector('#search-results');
 const resultsLiveRegion = document.querySelector('#search-results-announcement');
 const searchInput = document.querySelector('#search');
 const searchClearBtn = document.querySelector('#search__clear-btn');
+let activeIndex = -1;
+let searchQuery;
+
 
 //-----------------------------------------------------------------------------
 // Helpers
@@ -66,6 +69,7 @@ function displaySearchResults(results) {
         list.classList.add('search-results__list');
         resultsElement.append(list);
         resultsElement.setAttribute('data-results', 'true');
+        activeIndex = -1;
 
         for (const result of results) {
             const listItem = document.createElement('li');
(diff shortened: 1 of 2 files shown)


kecrily (approved):
LGTM, but maybe also add a shortcut for focus to the search box?

src/assets/js/search.js
```diff
@@ -110,16 +140,46 @@ searchInput.addEventListener('keyup', function (e) {
...
+        case "Enter":
+            resultsElement.querySelector('li.selected a').click();
+            break;
+        case "Escape":
+            e.preventDefault();
+            clearSearchResults();
+            break;
+
+        default:
+            break;
```
harish-sethuraman, line 174:
```suggestion
```

harish-sethuraman:
> but maybe also add a shortcut for focus to the search box? 

Yea that would be nice as well. We could do a similar implementation like other docs site to focus searchbox?  We could use `⌘ + K` keybinding?

shanpriyan:
> LGTM, but maybe also add a shortcut for focus to the search box?

Yeah we can

kecrily:
> We could use `⌘ + K` keybinding?

This is great, when I don't know the key bindings I always try `⌘ + K` first

shanpriyan:
> > We could use `⌘ + K` keybinding?
> 
> This is great, when I don't know the key bindings I always try `⌘ + K` first

Shall I add `⌘ + K` key binding to focus the search field?

harish-sethuraman:
Yep Go ahead. Use `⌘ + K` for mac and relevant key binding for windows as well. Please refer to other documentation site bindings as well like beta.reactjs.org, babeljs.io

src/assets/js/search.js
```diff
@@ -88,6 +92,30 @@ function displaySearchResults(results) {
...
+    const { offsetHeight, offsetTop } = activeElement;
+    const { offsetHeight: parentOffsetHeight, scrollTop } = scrollParent;
+
+    const isAbove = offsetTop < scrollTop;
+    const isBelow = (offsetTop + offsetHeight) > (scrollTop + parentOffsetHeight);
+
+    if (isAbove) {
+        scrollParent.scrollTo(0, offsetTop);
+    }
+    else if (isBelow) {
```
amareshsm, line 112: Why we have else if here?  Not just else.

shanpriyan: It is required to check if the element is below and then we need to scroll, removing that will change the scollbar position for visible elements also

nzakas:
This is looking really good! Just a few things:

1. When the up/down arrow is pressed, it should shift focus from the textbox to the link in the dropdown (call `focus()`). Right now, the focus remains in the textbox. The Esc key should then `focus()` back to the textbox.
2. When there are no search results, the Esc key doesn't close the dropdown. It should still close the dropdown.

amareshsm (approved):
LGTM

harish-sethuraman (approved):
LGTM

harish-sethuraman:
When I press down arrow from textbox and come down Im not able to move to textbox again with up arrow. Can we fix this? Other that this all LGTM

harish-sethuraman (commented):
When I press down arrow from textbox and come down Im not able to move to textbox again with up arrow. Can we fix this? Other that this all LGTM

nzakas (approved):
LGTM. Thanks!

PR #327, opened 2022-08-22
feat: make card cover can't reachable via `Tab`

kecrily:
Previously both cover and title could be reachable via keyboard navigation, but in fact both pointed to the same link. This is not accessibility friendly. Let's focus only on the more important title.

image
---

| Before | After |
| - | - |
| before | after |

### DIFF
--- a/src/_includes/components/card.macro.html
+++ b/src/_includes/components/card.macro.html
@@ -22,7 +22,7 @@
     {%- set img = ['/assets/images/blog-covers/', imageName] | join  -%}
     {%- set image_path = ["/blog-covers/", imageName] | join -%}
     {%- set image_alt = params.title -%}
-    <a href="{{ params.url }}" class="card__cover">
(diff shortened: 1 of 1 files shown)


amareshsm (approved):
LGTM

nzakas:
We actually hired an accessibility expert when putting this together and this was the setup they recommended. Why do you think this is a problem?

kecrily:
> Why do you think this is a problem?

Because cover and title are both `<a>` tags and point to the same link. If people use `Tab` for navigation, then if they want to focus on the second post and enter it, they need to press Tab twice to jump to the second post. First the cover of the first post, then the title of the first post, and finally focus the cover of the second post. It goes without saying that this is repetitive. 

The cover is of little significance to the visually impaired, as screen readers will only see cover and title as two similar elements.

nzakas:
Ah I see, thank you for explaining.

PR #328, opened 2022-08-22
ci: install playground dependencies

amareshsm:
**Description:** Install playground dependencies in ci.


.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         run: npm ci
       - name: Install playground dependencies
         working-directory: src/playground
-        run: npm ci
+        run: npm install
```
harish-sethuraman, line 24: This is expected. npm ci will delete any existing node_modules folder and relies on the package-lock. json file to install the specific version of each package. It is significantly faster than npm install because it skips some features.

.github/workflows/ci.yml
```diff
@@ -17,8 +17,11 @@ jobs:
         with:
           node-version: "16.x"
           cache: "npm"
-      - name: Install dependencies
+      - name: Install website dependencies
         run: npm ci
+      - name: Install playground dependencies
+        working-directory: src/playground
+        run: npm install
```
harish-sethuraman, line 24:
```suggestion
        run: npm ci
```

package.json
```diff
@@ -27,7 +27,7 @@
         "fetch": "npm-run-all --parallel fetch:*",
         "start": "npm-run-all build:sass --parallel watch:*",
         "build": "npm-run-all install:playground build:sass build:eleventy build:webpack images",
-        "lint": "eslint --ext=.js,.jsx . && markdownlint \"**/*.md\" -i node_modules",
+        "lint": "eslint --ext=.js,.jsx . && markdownlint \"**/*.md\" ",
```
harish-sethuraman, line 30: How about running lint parallely using `npm-run-all`? we can save a little bit as they lint different files?

.github/workflows/ci.yml
amareshsm: `npm ci` works only if we have `package-lock.json`.

.github/workflows/ci.yml
amareshsm: `npm ci` works only if we have `package-lock.json`.

package.json
amareshsm: yes I will try.

.github/workflows/ci.yml
snitin315: We already have an `install:playground` script which we can use.

https://github.com/eslint/eslint.org/blob/38d7f6cdd4651ae0077f6611b06ab2d3d8cf04bf/package.json#L18

package.json
snitin315: We should ignore `node_modeules`, why did we remove this?

.github/workflows/ci.yml
amareshsm: sure

snitin315 (approved):
LGTM, thanks!

PR #329, opened 2022-08-22
chore: remove duplicate style properties

amareshsm:
**Description:** Remove duplicate style properties.

### DIFF
--- a/src/assets/scss/animation.scss
+++ b/src/assets/scss/animation.scss
@@ -84,8 +84,6 @@
   }
 
   &__cursor {
-    width: 40px;
-    height: 40px;
     width: clamp(30px, 5vw, 50px);
     height: clamp(30px, 5vw, 50px);
     border-radius: 100%;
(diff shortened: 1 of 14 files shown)


nzakas (commented):
As I noted in the PR on the eslint repo, these duplications are intentional so we can support browsers that have CSS variables as well as those that do not.

kecrily:
Consider that ESLint and its website are geared towards front-end developers. At the same time, according to caniuse data, all evergreen browsers already support CSS variables[^1]. I don't think there is a need for compatibility. Compatibility never ends.

[^1]: [CSS Environment  Variables env()](https://caniuse.com/?search=css%20variable)

nzakas:
It’s true that compatibility never ends, however, good quality software aims to reach the largest audience possible. Having fallback styles allows us to do that.

We went through the trouble to do this when we built the site, so I don’t see a good argument to remove it.

nzakas:
Since we already have the duplicate properties for backwards compatibility, let's leave them. We don't need to add any more, though.

PR #339, opened 2022-09-04
feat: blog: How to create a custom ESLint rule

mike-solomon:
This blog post teaches people who are relatively new to ESLint how
to make their own custom rules. It walks through the entire process -
starting from thinking about the rule to coding each of the files.


src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,280 @@
...
+More details about all of the flags and options can be found in the [rule-basics](https://eslint.org/docs/latest/developer-guide/working-with-rules#rule-basics) section of the ESLint docs.
+
+For now, let's briefly talk through each of these pieces. I'll start by showing you some code, and then I'll talk through what each line does (you can find the gist for entire file [here](https://gist.github.com/mike-solomon/1f191393b5bef696937409491c8c559b) if that's easier). 
+
+#### Meta Section
+
+For the bacon example, the `meta` section would look like:
+
+{% highlight "javascript" %}
+{% raw %}
```
mike-solomon, line 198: I had to do highlighting like this a few times in this PR as when I tried the normal triple backticks, it stopped updating the site and threw parsing errors.

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,280 @@
...
+      }
+    }
+  };
+},
+{% endraw %}
+{% endhighlight %}
+
+All rules need to have a `create(context)` in them. What's inside of there, though, is where the differences lie. Let's talk through a few key pieces:
+
+* The `create(context)` function needs to return an object that ESLint calls to "visit" each of the nodes while traversing the tree. The simplest rules, like our bacon one, return an object that specifies a specific type of node and then runs some if statement s to check whether or not the rule is violated. More advanced rules will provide many different types of nodes and functions in combination.
```
amareshsm, line 254:
```suggestion
* The `create(context)` function needs to return an object that ESLint calls to "visit" each of the nodes while traversing the tree. The simplest rules, like our bacon one, return an object that specifies a specific type of node and then runs some if statements to check whether or not the rule is violated. More advanced rules will provide many different types of nodes and functions in combination.
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
amareshsm: because of this `{{identifier}}` template it is breaking I guess.

amareshsm:
[Lint check](https://github.com/eslint/eslint.org/runs/8179125904?check_suite_focus=true) is failing. Pls check the error and fix it.

mike-solomon:
@amareshsm Thanks for the note! Two things:

1. Linting appears to be broken for me locally. When I tried to run `npm run lint` it has a bunch of errors as it looks in the `src/playground` directory. I had to add `-i src/playground` to the end of the linting file in my `package.json` to get that to work. I didn't check that code in here as it felt like that should be separate from this. Maybe there's some other way I'm supposed to do linting locally? 
2. I was unable to figure out how to resolve `MD051 - Link fragments should be valid` - so I removed the links that linked some text to headers in the file. I do think that makes the doc worse but I wasn't sure how to fix it and I couldn't find any examples in my brief search around other posts. Do you know if that's something that's been done before?

Thanks!

amareshsm:
1. Linting appears to be broken for me locally. When I tried to run `npm run lint` it had a bunch of errors as it looked in the `src/playground` directory. I had to add `-i src/playground` to the end of the linting command in my `package.json` to get that to work. I didn't check that code in here as it felt like that should be separate from this. Maybe there's some other way I'm supposed to do linting locally? - **I guess what you did would work. This issue is fixed in this [PR]**(https://github.com/eslint/eslint.org/pull/328).
2. I was unable to figure out how to resolve `MD051 - Link fragments should be valid` - so I removed the links that linked some text to headers in the file. I do think that makes the doc worse but I wasn't sure how to fix it and I couldn't find any examples in my brief search around other posts. Do you know if that's something that's been done before? - **It is because the heading contains a special character `?`.**

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+
+There are four main places that new rules are commonly made:
+
+1. In the core [ESLint repository](https://github.com/eslint/eslint/tree/main/lib/rules) 
+    * *For rules that are useful to all JavaScript projects*
+2. In a plugin that already exists 
+    * *For rules that are useful to a specific subset of JavaScript projects*
+3. In a new plugin 
+    * *For rules that are useful to a specific subset of JavaScript projects, but that don't fit into any existing plugin*
+4. Locally in a specific project by utilizing the plugin [eslint-plugin-local-rules](https://www.npmjs.com/package/eslint-plugin-local-rules)
```
nzakas, line 80: You don't actually need to use a plugin to use a local rule. You can use the deprecated `--rulesdir`, but also, you can create a local plugin and reference it in your dependencies using the [`file:` prefix](https://stackoverflow.com/questions/14381898/local-dependency-in-package-json/14387210#14387210).

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+
+I won't go into more detail about what plugins are or how to make them in this article. Instead, I'll focus on showing you how to make a rule in either the core ESLint repository or in a plugin that already exists. If you want to learn more about how to make plugins, though, please check out the [Working with Plugins](https://eslint.org/docs/latest/developer-guide/working-with-plugins) guide on the ESLint website. 
+
+## How does ESLint parse code?
+
+The last thing you need to think about before coding your rule is the topic of parsing code. It's easy for us, as humans, to think, "Oh, I don't want any identifiers with bacon in them." Computers, unfortunately, are not that simple. They need ways to programmatically define what is an identifier vs. a function declaration vs. a literal and so on. 
+
+Fret not, though, as most of this work has already been done for us. ESLint uses something called [ESTree](https://github.com/estree/estree) to parse code. You can think of it as a way of breaking up code into a giant tree that ESLint can then iterate over to check things.
```
nzakas, line 91: Technically, ESTree is the AST format that ESLint uses, so it can't parse code.

ESLint uses a parser (called Espree) that parses JavaScript code into an ESTree-compatble AST format.

So Espree is the parser, ESTree is the AST format, and ESLint is the tool that uses Espree to create the AST. (Though ESLint can also use other parsers like babel/eslint-parser, Espree is the one it ships with.)

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+
+Fret not, though, as most of this work has already been done for us. ESLint uses something called [ESTree](https://github.com/estree/estree) to parse code. You can think of it as a way of breaking up code into a giant tree that ESLint can then iterate over to check things. 
+
+There's some [extensive documentation](https://web.archive.org/web/20210314002546/https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API#node_objects) you can check out that shows all the different types of nodes that code can be broken up into -- but an even easier way is to use the [AST Explorer](https://astexplorer.net/).
+
+The [AST Explorer](https://astexplorer.net/) lets you paste in JavaScript code and then shows you exactly what this tree looks like. You can click on any particular piece of code and you'll see the keys and values for it on the right side of the page.
```
nzakas, line 95:
```suggestion
[AST Explorer](https://astexplorer.net/) lets you paste in JavaScript code and then shows you exactly what this tree looks like. You can click on any particular piece of code and you'll see the keys and values for it on the right side of the page.
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+
+For the bacon example, let's copy over two lines to take a look at:
+
+```javascript
+let pork = "Meow";
+function bacon() {};
+```
+
+If you click on `let`, you should see a `VariableDeclaration` section highlighted on the right side. Likewise, if you click on `function`, you should see a `FunctionDeclaration` highlighted. But what if you click on `pork` or `bacon`? Fortunately for us, both of those cases show that `Identifier` is highlighted (See Figure 1).
```
nzakas, line 104:
```suggestion
If you click on `let` in the code section of AST Explorer, you should see a `VariableDeclaration` section highlighted on the right side. Likewise, if you click on `function`, you should see a `FunctionDeclaration` highlighted. But what if you click on `pork` or `bacon`? Fortunately for us, both of those cases show that `Identifier` is highlighted (See Figure 1).
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+
+## How does ESLint parse code?
+
+The last thing you need to think about before coding your rule is the topic of parsing code. It's easy for us, as humans, to think, "Oh, I don't want any identifiers with bacon in them." Computers, unfortunately, are not that simple. They need ways to programmatically define what is an identifier vs. a function declaration vs. a literal and so on. 
+
+Fret not, though, as most of this work has already been done for us. ESLint uses something called [ESTree](https://github.com/estree/estree) to parse code. You can think of it as a way of breaking up code into a giant tree that ESLint can then iterate over to check things. 
+
+There's some [extensive documentation](https://web.archive.org/web/20210314002546/https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API#node_objects) you can check out that shows all the different types of nodes that code can be broken up into -- but an even easier way is to use the [AST Explorer](https://astexplorer.net/).
```
nzakas, line 93: We should use the spec at https://github.com/estree/estree for this.

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+If you click on `let`, you should see a `VariableDeclaration` section highlighted on the right side. Likewise, if you click on `function`, you should see a `FunctionDeclaration` highlighted. But what if you click on `pork` or `bacon`? Fortunately for us, both of those cases show that `Identifier` is highlighted (See Figure 1).
+
+![Figure 1: AST Identifier Example](/assets/images/blog/2022/eslint-ast-example.png)
+*Figure 1: What an identifier in an Abstract Syntax Tree looks like*
+
+This piece is exactly what you'll need to begin coding the bacon rule. You can see that if you find an `Identifier`, you can check for the `name` and see if it violates the bacon restrictions. 
+
+If you visit the [Parser API](https://web.archive.org/web/20210314002546/https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API#miscellaneous) docs, you can see that `Identifiers` do indeed have a `name` field that is a `string` -- which matches what we saw above.
```
nzakas, line 111: Again, let's refer to the ESTree spec here.

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+
+Take some time to explore what code translates to what nodes by copying / pasting in various chunks of code that relate to the rule you're trying to make. This will make it easier for you when you go to write the code for the rule.
+
+## Generate the files
+
+It's now time to go about actually making your rule. (Yay!) Each rule in ESLint has three files named with its identifier. 
+
+*Note*: An identifier is a short, high-level description of the rule separated by dashes.
```
nzakas, line 119: I think you can pull this in as the last sentence of the previous paragraph.

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+
+For the bacon example, the identifier could be something like `no-bacon`. In which case the three files would be:
+
+1. `tests/lib/rules/no-bacon.js` -- A test file where you will test the rule to make sure it works as expected. I'd recommend you begin by writing this before you write your source file.
+2. `lib/rules/no-bacon.js` -- The source file where you will write the core logic of the rule.
+3. `docs/rules/no-bacon.md` -- A Markdown file that explains what your rule does. It largely will contain the answers to the questions discussed above.
+
+**Before examining each file, there's a few important links you should be aware of**:
```
nzakas, line 127:
```suggestion
**Before examining each file, there are a few important links you should be aware of**:
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+
+* ESLint provides a [Working with Rules](https://eslint.org/docs/latest/developer-guide/working-with-rules) guide on their website that goes into a lot of detail about each of the files and what every property is. It's something you'll definitely want to check out when you start making your own rules.
```
nzakas, line 129:
```suggestion
* ESLint provides a [Working with Rules](https://eslint.org/docs/latest/developer-guide/working-with-rules) guide that goes into a lot of detail about each of the files and what every property is. It's something you'll definitely want to check out when you start making your own rules.
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+To begin your test file, you'll want to import your rule and the ESLint `RuleTester`:
+
+```javascript
+const rule = require("../../../lib/rules/no-bacon");
+const RuleTester = require("eslint").RuleTester;
+```
+
+From there, you'll want to create your own `ruleTester` object that you can use to run your test cases. You can provide it with options that will act as defaults for all the test cases. 
+
+For instance, the `RuleTester`, by default, uses [ECMA Version 5](https://www.w3schools.com/js/js_es5.asp). If you're used to modern JavaScript, some of your test cases might not work. To fix that, you can specify `parserOptions` when you make this tester so that all your tests use ECMA Version 6 instead:
```
nzakas, line 145:
```suggestion
For instance, the `RuleTester`, by default, uses ECMAScript 5. If you're used to modern JavaScript, some of your test cases might not work. To fix that, you can specify `parserOptions` when you make this tester so that all your tests use ECMAScript 6 instead by using the `ecmaVersion` option:
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
nzakas: Yeah, I believe you can use triple backticks and just use the raw block around `{{identifier}}`.

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+* The `messages` key (line 12) is a way of providing one or more warnings / errors for the user. By defining these here, it makes it so you don't have to copy the same message to many different files or locations. For instance, the tests I wrote above only have to reference the key of `avoidBacon` instead of copying the entire message to that file.
+* In the `avoidBacon` message (line 13), you can see that I have `identifier` surrounded by double brackets. By enclosing a variable name in double brackets like this, I'm telling ESLint that this warning message expects a variable called `identifier` to be passed in. I'll need to add that in the context section below. You can use this to provide better error messages to your users.
+
+If you want to learn more about what flags and settings exist, please read the  [ESLint Rule Basics](https://eslint.org/docs/latest/developer-guide/working-with-rules#rule-basics) guide.
+
+#### Context Function
```
nzakas, line 233: You are using "Context function" throughout, but it should be "the context argument" or "the context object."

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,286 @@
...
+        });
+      }
+    }
+  };
+},
+```
+
+All rules need to have a `create(context)` in them. What's inside of there, though, is where the differences lie. Let's talk through a few key pieces:
+
+* The `create(context)` function needs to return an object that ESLint calls to "visit" each of the nodes while traversing the tree. The simplest rules, like our bacon one, return an object that specifies a specific type of node and then runs some if statements to check whether or not the rule is violated. More advanced rules will provide many different types of nodes and functions in combination.
```
nzakas, line 259: Technically, the methods on the object are called, not the object itself.

nzakas (changes requested):
This is looking really good, thank you! I left a few comments throughout but nothing major.

mike-solomon:
@nzakas Thanks so much for all the suggestions! I believe I've fixed everything you suggested. I would appreciate if you glance over the `create(context)` section and the `EStree` section again as I'm not 100% confident I explained them correctly.

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -88,11 +88,11 @@ I won't go into more detail about what plugins are or how to make them in this a
...
 
-There's some [extensive documentation](https://web.archive.org/web/20210314002546/https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API#node_objects) you can check out that shows all the different types of nodes that code can be broken up into -- but an even easier way is to use the [AST Explorer](https://astexplorer.net/).
+There's some [extensive documentation](https://github.com/estree/estree#ast-descriptor-syntax) you can check out that shows all the different types of nodes that code can be broken up into -- but an even easier way is to use the [AST Explorer](https://astexplorer.net/).
 
-The [AST Explorer](https://astexplorer.net/) lets you paste in JavaScript code and then shows you exactly what this tree looks like. You can click on any particular piece of code and you'll see the keys and values for it on the right side of the page.
+[AST Explorer](https://astexplorer.net/) lets you paste in JavaScript code and then shows you exactly what this tree looks like. You can click on any particular piece of code and you'll see the keys and values for it on the right side of the page.
```
nzakas, line 95:
```suggestion
AST Explorer lets you paste in JavaScript code and then shows you exactly what this tree looks like. You can click on any particular piece of code and you'll see the keys and values for it on the right side of the page.
```

nzakas: Since we already have a link in the preceding sentence.

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -101,32 +101,30 @@ let pork = "Meow";
 function bacon() {};
 ```
 
-If you click on `let`, you should see a `VariableDeclaration` section highlighted on the right side. Likewise, if you click on `function`, you should see a `FunctionDeclaration` highlighted. But what if you click on `pork` or `bacon`? Fortunately for us, both of those cases show that `Identifier` is highlighted (See Figure 1).
+If you click on `let` in the code section of the AST explorer, you should see a `VariableDeclaration` section highlighted on the right side. Likewise, if you click on `function`, you should see a `FunctionDeclaration` highlighted. But what if you click on `pork` or `bacon`? Fortunately for us, both of those cases show that `Identifier` is highlighted (See Figure 1).
```
nzakas, line 104:
```suggestion
If you click on `let` in the code section of AST Explorer, you should see a `VariableDeclaration` section highlighted on the right side. Likewise, if you click on `function`, you should see a `FunctionDeclaration` highlighted. But what if you click on `pork` or `bacon`? Fortunately for us, both of those cases show that `Identifier` is highlighted (See Figure 1).
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -101,32 +101,30 @@ let pork = "Meow";
...
 
 ![Figure 1: AST Identifier Example](/assets/images/blog/2022/eslint-ast-example.png)
 *Figure 1: What an identifier in an Abstract Syntax Tree looks like*
 
 This piece is exactly what you'll need to begin coding the bacon rule. You can see that if you find an `Identifier`, you can check for the `name` and see if it violates the bacon restrictions. 
 
-If you visit the [Parser API](https://web.archive.org/web/20210314002546/https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API#miscellaneous) docs, you can see that `Identifiers` do indeed have a `name` field that is a `string` -- which matches what we saw above.
+If you visit the [ES5 Spec](https://github.com/estree/estree/blob/master/es5.md#identifier), you can see that `Identifiers` do indeed have a `name` field that is a `string` -- which matches what we saw above.
```
nzakas, line 111:
```suggestion
If you visit the [ESTree Specification](https://github.com/estree/estree/blob/master/es5.md#identifier), you can see that `Identifiers` do indeed have a `name` field that is a `string` -- which matches what we saw above.
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -101,32 +101,30 @@ let pork = "Meow";
...
 
 Take some time to explore what code translates to what nodes by copying / pasting in various chunks of code that relate to the rule you're trying to make. This will make it easier for you when you go to write the code for the rule.
 
 ## Generate the files
 
-It's now time to go about actually making your rule. (Yay!) Each rule in ESLint has three files named with its identifier. 
-
-*Note*: An identifier is a short, high-level description of the rule separated by dashes.
+It's now time to go about actually making your rule. (Yay!) Each rule in ESLint has three files named with its identifier. *Note*: An identifier is a short, high-level description of the rule separated by dashes.
```
nzakas, line 117: We can just let this be part of the paragraph.

```suggestion
It's now time to go about actually making your rule. (Yay!) Each rule in ESLint has three files named with its identifier. A rule identifier is a short, high-level description of the rule in dash-case, such as `prefer-const`.
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,282 @@
...
+* In the `avoidBacon` message (line 13), you can see that I have `identifier` surrounded by double brackets. By enclosing a variable name in double brackets like this, I'm telling ESLint that this warning message expects a variable called `identifier` to be passed in. I'll need to add that in the context section below. You can use this to provide better error messages to your users.
+
+If you want to learn more about what flags and settings exist, please read the  [ESLint Rule Basics](https://eslint.org/docs/latest/developer-guide/working-with-rules#rule-basics) guide.
+
+#### Context Object
+
+For the bacon example, the `context` object would look like: 
+
+```javascript
+create(context) {
```
nzakas, line 234: This is technically invalid JavaScript syntax. Can you flesh out this section by doing something like:

```js
module.exports = {
    // meta omitted for clarity
   create(context) {
      ...
    }

};
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,282 @@
...
+More details about all of the flags and options can be found in the [rule-basics](https://eslint.org/docs/latest/developer-guide/working-with-rules#rule-basics) section of the ESLint docs.
+
+For now, let's briefly talk through each of these pieces. I'll start by showing you some code, and then I'll talk through what each line does (you can find the gist for entire file [here](https://gist.github.com/mike-solomon/1f191393b5bef696937409491c8c559b) if that's easier). 
+
+#### Meta Section
+
+For the bacon example, the `meta` section would look like:
+
+```javascript
+meta: {
```
nzakas, line 202: This is technically invalid JS syntax, which may confuse some readers. Can we build this out a bit?

```js
module.exports = {
   meta: {
       ...
    },
   create(context) {
      // omitted for clarity
    }

};
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,282 @@
...
+If you want to learn more about what flags and settings exist, please read the  [ESLint Rule Basics](https://eslint.org/docs/latest/developer-guide/working-with-rules#rule-basics) guide.
+
+#### Context Object
+
+For the bacon example, the `context` object would look like: 
+
+```javascript
+create(context) {
+  return {
+    Identifier(node) {
```
nzakas, line 236: One thing to keep in mind with this approach is that it will also flag something like this:

```js
let a = bacon;
```

In this case, you're not creating an identifier named `bacon`, but rather consuming one that you may not be able to change.

Can you add a note calling this out?

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,282 @@
...
+            identifier
+          }
+        });
+      }
+    }
+  };
+},
+```
+
+All rules need to have a `create(context)` in them. What's inside of there, though, is where the differences lie. Let's talk through a few key pieces:
```
nzakas, line 253: What we're really talking about is having a particular method.
```suggestion
All rules need to have a `create()` method. What's inside of the method, though, is where the differences lie. Let's talk through a few key pieces:
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,282 @@
...
+
+* The `create(context)` object is an object that ESLint uses when it "visits" each of the nodes while traversing the tree. The simplest rules, like our bacon one, specify a specific type of node and then run some if statements to check whether or not the rule is violated. More advanced rules will provide many different types of nodes and functions in combination.
```
nzakas, line 255:
```suggestion
* The object returned from the `create()` method tells ESLint which nodes the rule is interested in. When ESLint traverses an AST, it stops on the specified nodes to run the code from your rule. The simplest rules, like our bacon one, specify one specific type of node and then run some `if` statements to check whether or not the rule is violated. More advanced rules will provide many different types of nodes and functions in combination.
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,282 @@
...
+  * Remember earlier how you used the AST Explorer to get a better understanding of how code is broken down? You'll now put that to use here by typing something like: `Identifier(node) {}` and then adding some code for parsing whatever object you went with inside of the brackets.
```
nzakas, line 256: I'm not quite sure what you're saying here. "Creating something like" is a bit confusing because we are using a specific example in the post. Also the phrase "parsing whatever object" is incorrect (you can't parse objects, you can parse strings), so I'm not entirely sure what it is you're trying to say here. Do you mean that you can look at the node that was passed in to determine if your rule passes?

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,282 @@
...
+* `context.report()` is how you let ESLint know that something violated the rule. In there, you'll want to provide a `node` object and a `messageId` that has the same key as the one in the `meta` section. You can also provide extra data so the user has a better idea of what went wrong (such as by showing them the identifier that has bacon in it).
```
nzakas, line 257: "In there" is a bit vague.

```suggestion
* `context.report()` is how you let ESLint know that something violated the rule. When you call `context.report()`, you'll want to provide a `node` object and a `messageId` that has the same key as the one in the `meta` section. You can also provide extra data so the user has a better idea of what went wrong (such as by showing them the identifier that has bacon in it).
```

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,282 @@
...
+
+If you want a more detailed look into everything around this `context` object, please read the [ESLint Context Object docs](https://eslint.org/docs/latest/developer-guide/working-with-rules#the-context-object).
```
nzakas, line 259:
```suggestion
If you want a more detailed look into everything around the `context` object, please read the [ESLint Context Object docs](https://eslint.org/docs/latest/developer-guide/working-with-rules#the-context-object).
```

nzakas (changes requested):
Thanks for the edits. I recommended some clarifications throughout, particularly around the `create()` method and the `context` object. Let me know if these make sense.

src/content/drafts/how-to-create-a-custom-eslint-rule.md
mike-solomon: Ooh that's a great call. Thanks!

src/content/drafts/how-to-create-a-custom-eslint-rule.md
mike-solomon: Apologies for the confusion. My intention was to call out that the `Identifier(node) {}` is something that would be changed depending on the rule. So if someone was wanting a rule about functions, they wouldn't use the `Identifier` part of things.

I'll re-write the section to try and make it clearer.

mike-solomon:
I think I fixed the latest round of suggestions @nzakas . Thanks again for taking the time to review these :)

nzakas:
Thanks so much! We will take some time to review this and get back to you soon.

src/content/drafts/how-to-create-a-custom-eslint-rule.md
```diff
@@ -0,0 +1,293 @@
...
+Now that you have some idea of how your rule will work, you should take some time to think about who would benefit from this rule. Some questions to ask yourself are:
+
+* Is this a rule that would benefit just your company or is it something that other people might want?
+* If this rule **is** specific to your company:
+  * Is it specific to a single project?
+  * Is it something that many different projects could benefit from?
+* If this rule **isn't** specific to your company:
+  * Is it something that benefits *all* JavaScript projects or is it more specific than that? For example, a rule that provides suggestions for TypeScript apps that use GitHub would be something that is beneficial for some projects, whereas a rule that suggests to not use `console.log` would be beneficial for all types of projects. 
+
+Once you have those answers, you can then think aboutv where the code for this rule will live.
```
harish-sethuraman, line 68:
```suggestion
Once you have those answers, you can then think about where the code for this rule will live.
```

nzakas:
Thanks so much for this. We will follow up over email.

PR #344, opened 2022-09-12
fix: wrong or duplicate focusable

kecrily:
1. It is pointless to focus on these two places, they are not clickable.

| 1 | 2 |
| - | - |
| 截屏2022-09-12 13 32 37 | 截屏2022-09-12 13 32 11 |

2. There are places where a class name called `focusable` exists, but there is no such style.
3. Some svg's have two identical `focusable=false` attributes.


src/_includes/components/testimonials-slider.html
```diff
@@ -2,7 +2,7 @@
     <div class="c-slider__slides-container" data-slides>
         <div class="c-slider__slides-wrapper" data-slides-wrapper>
             {%- for item in site.donate_page.testimonials.items -%}
-            <div class="c-slider__slide focusable" data-slide>
+            <div class="c-slider__slide" data-slide>
```
harish-sethuraman, line 5: Screenshot 2022-09-14 at 9 49 41 PM
I still see the focus in your deploy preview can you check this?  or is this expected?

kecrily: I left it out. I'm not sure which node is affecting this behavior right now, and I'm going to wake up tomorrow and check to see.

nzakas:
I think the first two might be intentional for easier keyboard navigation. Let me check with Sara.

SaraSoueidan:
@kecrily Hey Percy!

Thanks so much for contributing to the a11y of the site!

**Re focusability & clickability**:

In general: it doesn't have to be clickable to be focusable. For example, scrollable regions must be focusable even though they're not "clickable" per se. Sometimes making a group focusable is useful for screen reader announcements. 

1. For the slider: focusing on the slide allows the screen reader to announce its contents without the user needing to navigate to it. This is sometimes helpful, particularly when there are no interactive elements within the slide. So this focusability is targeting screen reader users, not sighted keyboard or mouse users. That being said, it is not _a requirement_ to keep it focusable. So I’ll leave this up to @nzakas to decide here. I have no strong opinion on it.
2. Re list of donations: Excellent catch. I had the `tabindex="0"` set on it because the design had the list of donations scroll horizontally on mobile. We (@nzakas and I) eventually changed the design to make it more mobile-friendly. I must have forgotten to remove it (obviously). Great catch, thank you!

**Re duplicate `focusable="false"` on `<svg>`s**: Another great catch, thanks!

> There are places where a class name called focusable exists, but there is no such style.

The `:focus-visible` polyfill will add this class name to elements that are focusable. I'm not sure which elements have this class name but have no focus styles. I'm thinking it's either something I have missed, or something I have deliberately _not_ added focus styles to (for some reason; I'll need to know which elements have it to tell.)

src/_includes/components/testimonials-slider.html
```diff
@@ -1,8 +1,8 @@
 <div class="c-slider c-slider--testimonials" data-slider data-aria-label="Testimonials">
     <div class="c-slider__slides-container" data-slides>
-        <div class="c-slider__slides-wrapper" data-slides-wrapper>
+        <div class="c-slider__slides-wrapper" tabindex="-1" data-slides-wrapper>
```
nzakas, line 3: @kecrily based on @SaraSoueidan's comments, can we revert these changes in the slider?

nzakas (changes requested):

nzakas (approved):
LGTM. Thanks!

PR #345, opened 2022-09-12
fix: remove duplicate and unnecessary `role` aria

kecrily:
The `ul` element has an implicit `list` role attribute.


harish-sethuraman (commented):
Why'd we want this package? we already don't have one 😅

kecrily:
I don't understand what you are saying? Do you mean package lock.json for playground?

harish-sethuraman:
yes. Playground folder did not have a lock file before and seems like it got added in PR. We'd really want that?

nzakas:
Safari has a bug where it removes the role of “list” if list-type is “none”, which is why we have added role=“list” and it shouldn’t be removed.

further reading:
https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html

kecrily:
Is the use of two roles here also intentional?

https://github.com/eslint/eslint.org/pull/345/files#diff-f826482eabee5663501fbf4bb9acfdbec0e7e3b231250cb98c017e27e11bff28

nzakas:
Duplicates are a mistake. All lists should be role=list

PR #346, opened 2022-09-12
feat: add the missing `title` attribute of link

kecrily:


nzakas:
Thanks for the suggestion. The current best practices for accessibility actually say not to use the title attribute in most places. Here is some further reading:

https://www.24a11y.com/2017/the-trials-and-tribulations-of-the-title-attribute/

https://www.a11yproject.com/posts/title-attributes/

kecrily:
I took the advice of [w3 validtor](https://validator.w3.org), but the content of these two articles was very helpful. I guess I should take a screen reader and test it out for real.

nzakas:
I know I mentioned before, but we did hire an accessibility export to implement the new website so we would have all of the latest accessibility best practices implemented. Some of the best practices are counter-intuitive, for sure. We did accessibility testing as part of the launch too.

PR #348, opened 2022-09-12
feat: add feed link in `<head>`

kecrily:


nzakas (approved):
Thanks for adding this.

PR #352, opened 2022-09-14
fix: update css selector for anchors

chenxsan:
This seems to be a mistake where anchors were generated for sidebar headings as well (those red ones).

I think it should only target the blog content, i.e., the green part.


kecrily (approved):
LGTM, thank you

harish-sethuraman (approved):
LGTM. Thanks

PR #357, opened 2022-10-13
feat: Add zh-hans docs

nzakas:
This enables individual sites to refer to their own documentation sites. zh-hans is the first locale that has its own documentation site, so that is the only one that is enabled.

### DIFF
--- a/src/_data/sites/zh-hans.yml
+++ b/src/_data/sites/zh-hans.yml
@@ -16,6 +16,9 @@ language:
   title: Simplified Chinese
 locale: zh-hans
 hostname: zh-hans.eslint.org
+locals:
+  docs: zh-hans-docs.netlify.app
+  blog: false
 
 #------------------------------------------------------------------------------
 # Analytics
--- a/src/_data/sites/zh-hans.yml
+++ b/src/_data/sites/zh-hans.yml
@@ -17,7 +17,8 @@ language:
 locale: zh-hans
 hostname: zh-hans.eslint.org
 locals:
-  docs: zh-hans-docs.netlify.app
+  docs_latest: zh-hans-docs.netlify.app
+  docs_head: false
   blog: false
 
 #------------------------------------------------------------------------------
--- a/src/static/redirects.njk
+++ b/src/static/redirects.njk
@@ -18,7 +18,7 @@ eleventyExcludeFromCollections: true
 # Internal Redirects
 /demo/*                             /play/:splat 302!
 
-{% if site.locals.docs %}
+{% if site.locals.docs_latest %}
 
 # Old-Style Docs to New-Style Docs
 /docs/rules/*                       /docs/latest/rules/:splat 301!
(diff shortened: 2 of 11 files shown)


src/static/redirects.njk
```diff
@@ -18,24 +18,25 @@ eleventyExcludeFromCollections: true
...
 /docs/user-guide/*                  /docs/latest/user-guide/:splat 301!
 /docs/maintainer-guide/*            /docs/latest/maintainer-guide/:splat 301!
 /docs/developer-guide/*             /docs/latest/developer-guide/:splat 301!
 
-# Version-Specific Docs
-/docs/head/*                        https://docs-eslint.netlify.app/:splat 200!
-/docs/latest/*                      https://latest--docs-eslint.netlify.app/:splat 200!
+# Regular Docs
+/docs/head/*                        https://{{ site.locals.docs }}/:splat 200!
+/docs/latest/*                      https://latest--{{ site.locals.docs }}/:splat 200!
```
mdjermanovic, line 31: We should create `latest` branch in https://github.com/eslint/zh-hans.docs.eslint.org and configure Netlify to deploy it?

nzakas: Yup, that was on my todo list.

Although, I think maybe these docs will only ever be translated from `latest`, so maybe we should always have `/docs/head` go to the English site?

kecrily: I think the Chinese translation should be `/docs/head` and `/docs/latest` should go to English site. Because Chinese can never be up to date, I sync each version as it is released.

mdjermanovic: > Although, I think maybe these docs will only ever be translated from `latest`, so maybe we should always have `/docs/head` go to the English site?

Makes sense to me :+1:

mdjermanovic: > I think the Chinese translation should be `/docs/head` and `/docs/latest` should go to English site. Because Chinese can never be up to date, I sync each version as it is released.

`/docs/head` relates to the latest development (i.e., the eslint/eslint `main` branch), and those docs may contain some features that haven't been released yet, so most users should rather be interested in the `/docs/latest` version. It would be confusing if `/docs/head` on different sites has a different meaning, like "the latest translated version".

`/docs/latest` should ideally be the latest published version with the `latest` tag. There will obviously be a little lag on translated sites, but the version selector box makes it clear which particular version the docs are describing.

nzakas: I agree. We should stick with HEAD meaning the latest unreleased stuff while “latest” is the most recent published release (which matches the npm definition as well).

src/_data/sites/zh-hans.yml
```diff
@@ -16,6 +16,10 @@ language:
   title: Simplified Chinese
 locale: zh-hans
 hostname: zh-hans.eslint.org
+locals:
+  docs_latest: zh-hans-docs.netlify.app
```
mdjermanovic, line 20: This version shows HEAD in the version selector, while what we want to show is the latest translated version (currently v8.25.0)?

https://deploy-preview-357--zh-hans-eslint.netlify.app/docs/latest/

image

Selecting v8.25.0 reloads the same page, and there's again HEAD in the version selector.

mdjermanovic: Oh, now I see https://github.com/eslint/zh-hans.docs.eslint.org/issues/76. For some reason, I haven't been getting notifications for the past few hours.

PR #365, opened 2022-11-04
chore: remove duplicate selectors and enable stylelint rule

harish-sethuraman:
similar to https://github.com/eslint/eslint/pull/16467/

### DIFF
--- a/src/assets/scss/foundations.scss
+++ b/src/assets/scss/foundations.scss
@@ -62,6 +62,7 @@ html {
 }
 
 body {
+    font-size: var(--step-0);
     position: relative;
     margin: 0 auto;
     line-height: 1.5;
@@ -284,11 +285,6 @@ ol {
     }
 }
 
-p:empty {
-    margin: 0;
-    display: none;
-}
-
 figure {
     margin-bottom: 4rem;
     margin-block-end: 4rem;
@@ -340,10 +336,6 @@ nav {
 }
 
 /* typography */
-body {
-    font-size: var(--step-0);
-    line-height: 1.5;
-}
 
 .eyebrow {
     color: var(--link-color);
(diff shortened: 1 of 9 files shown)


snitin315 (commented):
Looks good to me. I hope this is not intentional. I would like @nzakas to review it.

harish-sethuraman:
This is not auto fix but a manual one. I went through selectors and out them to a diff checker and removed common ones.

src/assets/scss/foundations.scss
```diff
@@ -284,11 +285,6 @@ ol {
     }
 }
 
-p:empty {
```
amareshsm, line 287: Is this removed intentionally?

harish-sethuraman: Yes sir. This is already there https://github.com/eslint/eslint.org/blob/337f2bc851f9361c7802d87f9b117aee974e11c7/src/assets/scss/foundations.scss#L169-L172

amareshsm (approved):
LGTM

src/assets/scss/tokens/typography.scss
```diff
@@ -1,20 +1,18 @@
 /* @link https://utopia.fyi/type/calculator?c=320,16,1.125,1280,16,1.25,6,2,&s=0.75|0.5|0.25,1.5|2|3|4|6,s-l */
 
-:root {
```
nzakas, line 3: I believe this needs to stay where it is due to the @media(screen) block below changing `--fluid-screen`. Otherwise, the order of overrides is backwards.

nzakas (changes requested):
I'm not sure about a lot of these changes because the change the order in which properties are assigned and overridden. This isn't just about making sure that selectors aren't duplicated, it's also about making sure the cascade is working correctly. There are a lot of places where variables and properties are being overridden by `@media` blocks later, so moving properties from below `@media` blocks to above or vice versa actually changes the meaning of the CSS.

Please go back through and double-check that the order for these properties is correct.

src/assets/scss/tokens/typography.scss
harish-sethuraman: This was a miss from my side. Will take time to iterate upon this and fix it

harish-sethuraman:
Closing in favour of https://github.com/eslint/eslint.org/pull/421

PR #367, opened 2022-11-06
fix: open rule docs in new tab from the Popup component

snitin315:
Open docs in a new tab. Because as a user I want to look at the code in the editor and docs back and forth.

Screenshot 2022-11-06 at 2 39 06 PM

Currently, this is also the behavior of links in the console.

Screenshot 2022-11-06 at 2 36 20 PM


kecrily (approved):
LGTM

PR #371, opened 2022-11-26
perf: debounce search

harish-sethuraman:
debounced the algolia search query to go once instead of every character

### DIFF
--- a/src/assets/js/search.js
+++ b/src/assets/js/search.js
@@ -114,6 +114,20 @@ function maintainScrollVisibility(activeElement, scrollParent) {
     }
     
 }
+function debounce(callback, delay) {
+    let timerId;
+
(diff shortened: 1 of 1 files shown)


src/assets/js/search.js
```diff
@@ -114,6 +114,20 @@ function maintainScrollVisibility(activeElement, scrollParent) {
     }
     
 }
+function debounce(callback, delay) {
+    let timerId;
+
+    return (...args) => {
+        clearTimeout(timerId);
+        timerId = setTimeout(() => {
+            callback(...args);
```
amareshsm, line 123:
```suggestion
            callback.apply(this, args)
```

amareshsm (approved):
LGTM

PR #381, opened 2022-12-10
feat: add mastodon link

amareshsm:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

#### What changes did you make? (Give an overview)
Added Mastodon link along with our other social buttons.
image
image

#### Related Issues
Resolves https://github.com/eslint/eslint.org/issues/378

PR #382, opened 2022-12-14
fix: social icons layout bug

amareshsm:
#### Prerequisites checklist

Fixed the layout bug - https://github.com/eslint/eslint.org/pull/381#issuecomment-1349514328.

| before | After |
|--------|------|
|  image | image

### DIFF
--- a/src/assets/scss/components/social-icons.scss
+++ b/src/assets/scss/components/social-icons.scss
@@ -10,7 +10,6 @@
         padding: 0;
         margin-left: -1rem;
         margin-inline-start: -1rem;
-        display: inline-flex;
 
         li {
             margin: 0;


harish-sethuraman (commented):
Let's fix this first at main ESLint docs and then lets make the same fix here.

Tanujkanti4441:
Screenshot 2022-12-15 002807

the language select is also going out of the screen in homepage.

src/assets/scss/components/social-icons.scss
```diff
@@ -10,7 +10,6 @@
         padding: 0;
         margin-left: -1rem;
         margin-inline-start: -1rem;
```
shanpriyan, line 12: if we do the following changes we don't need to wrap the mastodon icon , it appears in the same line even at different breakpoints.
<details>
<summary>Screenshot (1052 x 522)</summary>


image


</details>

```suggestion
        display: flex;
        justify-content: space-between;
```


Remove `padding` from here 

https://github.com/eslint/eslint.org/blob/885d0d939ebc77370f56db475c750893552335e4/src/assets/scss/components/social-icons.scss#L22

nzakas: How does that work on the homepage footer?

harish-sethuraman:
@Tanujkanti4441  I guess it is fixed in this PR's preview right?

Tanujkanti4441:
> 

ya that select one is fixed in this PR my bad.

problem is all about the wrap part

Screenshot 2022-12-15 223031

Tanujkanti4441:
Adding these lines of code to the [src/assets/scss/components/social-icons.scss](https://github.com/eslint/eslint.org/pull/382/files/e372b7b6ba393b27e85b0ee24dab5c875193265e#diff-4dd655cea29886ba0c299b787f030ac980a50b53d8c176e5b458894404a1eae5) file is solving the problem for me.
Screenshot 2022-12-15 233615


please tell me if is it ok to sharing the code by a screenshot rather than the code pen

src/assets/scss/components/social-icons.scss
shanpriyan: we need to change the properties little bit to achieve that.

shanpriyan: Added nowrap changes with socials layout fix in https://github.com/eslint/eslint.org/pull/386 (homepage and playground)

amareshsm:
> Adding these lines of code to the [src/assets/scss/components/social-icons.scss](https://github.com/eslint/eslint.org/pull/382/files/e372b7b6ba393b27e85b0ee24dab5c875193265e#diff-4dd655cea29886ba0c299b787f030ac980a50b53d8c176e5b458894404a1eae5) file is solving the problem for me.
> 
> Screenshot 2022-12-16 013348
> 
> please let me know if it is ok to sharing the code by a screenshot rather than the code pen

@Tanujkanti4441  Thanks for the suggestion. it's similar to https://github.com/eslint/eslint.org/pull/386.

harish-sethuraman:
Closing in favour of https://github.com/eslint/eslint.org/pull/386

PR #391, opened 2022-12-24
fix: update example code

chenxsan:
#### Prerequisites checklist

- [x] I have read the [contributing guidelines](https://github.com/eslint/eslint/blob/HEAD/CONTRIBUTING.md).

#### What is the purpose of this pull request?

Update the example for ESM per https://github.com/eslint/eslintrc/blob/main/README.md#usage

PR #392, opened 2022-12-24
fix: use sponsor's GitHub username when name is missing

chenxsan:
#### Prerequisites checklist

The CI [has failed](https://github.com/eslint/eslint.org/actions/runs/3770301500/jobs/6410053617) because of a `null` value of name coming from here https://github.com/eslint/eslint.org/commit/415e0bba84b8d205faa5c47924c928b0b0074dba#diff-ff108750031135fb05819cca15c41069cf217df9c9e21745012a5af2e965c55bR256.

And I can confirm it's not a glitch or something because the Web version of GitHub GraphQL API did return `null` for this sponsor:

CleanShot 2022-12-24 at 19 38 49@2x

1. Set a fallback value for `name` in template.

    Currently I'm setting it to `''`, but I'm not sure it's a good option as it could mess the layout:

    

    But it shouldn't be a problem once the data fetching is done next midnight because there's another fallback value in `fetch-sponsors.js`, see next.

2. Set the value of `name` to `login` when `null` in `fetch-sponsors.js`

### DIFF
--- a/src/content/pages/sponsors.html
+++ b/src/content/pages/sponsors.html
@@ -157,7 +157,7 @@ hook: "sponsors_page"
             {% for donation in donations | limitTo(50) %}
             <li class="donations-item">
                 {% set donation_image = donation.image %}
-                {% set donation_name = donation.name %}
+                {% set donation_name = donation.name or '' %}
                 {% set donation_amount = donation.amount | dollars %}
                 {% set donation_date = donation.date | readableDateFromISO %}
 
--- a/src/content/pages/sponsors.html
+++ b/src/content/pages/sponsors.html
@@ -157,7 +157,7 @@ hook: "sponsors_page"
             {% for donation in donations | limitTo(50) %}
             <li class="donations-item">
                 {% set donation_image = donation.image %}
-                {% set donation_name = donation.name or '' %}
+                {% set donation_name = donation.name %}
                 {% set donation_amount = donation.amount | dollars %}
                 {% set donation_date = donation.date | readableDateFromISO %}
 
--- a/tools/fetch-sponsors.js
+++ b/tools/fetch-sponsors.js
@@ -239,7 +239,7 @@ async function fetchGitHubSponsors() {
         .filter(transaction => transaction.tier && transaction.tier.isOneTime)
         .map(({ sponsor, timestamp, tier, id }) => ({
             id,
-            name: sponsor.name,
+            name: sponsor.name || sponsor.login,
             image: sponsor.avatarUrl,
             url: sponsor.websiteUrl || sponsor.url,
             amount: tier.monthlyPriceInDollars,
--- a/tools/fetch-sponsors.js
+++ b/tools/fetch-sponsors.js
@@ -227,7 +227,7 @@ async function fetchGitHubSponsors() {
     // return an array in the same format as Open Collective
     const sponsors = organization.sponsorshipsAsMaintainer.nodes
         .map(({ sponsor, tier }) => ({
-            name: sponsor.name,
+            name: sponsor.name || sponsor.login,
             image: sponsor.avatarUrl,
             url: sponsor.websiteUrl || sponsor.url,
             monthlyDonation: tier.monthlyPriceInDollars,
@@ -249,7 +249,7 @@ async function fetchGitHubSponsors() {
         .filter(donation => {
             // invoiced recurring donations are incorrectly marked as one-time
             const sponsor = sponsors.find(sponsor => sponsor.name === donation.name);
-            
+
             // only include if the amount is different than the monthly amount
             return sponsor ? sponsor.monthlyDonation !== donation.amount : true;
         });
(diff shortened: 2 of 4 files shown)


chenxsan:
As you can see from the actions, CI no longer fails like https://github.com/eslint/eslint.org/pull/391

src/content/pages/sponsors.html
```diff
@@ -157,7 +157,7 @@ <h2 class="section-title h3" id="donations-label">{{ site.sponsors_page.recent_d
             {% for donation in donations | limitTo(50) %}
             <li class="donations-item">
                 {% set donation_image = donation.image %}
-                {% set donation_name = donation.name %}
+                {% set donation_name = donation.name or '' %}
```
chenxsan, line 160: See https://mozilla.github.io/nunjucks/templating.html#default-value-default-boolean

kecrily: If `donation.name` never seems to go to `null` when the data is re-fetched, then it seems that we don't need to set a fallback and just wait patiently for the data to update

chenxsan: Nah, it will be `null` unless the user sets the name field. It's easy to reproduce, remove your name from your github profile then fetch data with the github graphql explorer:

kecrily: https://github.com/eslint/eslint.org/pull/392/files#diff-e459d8973a8d888faaadd76b2c93a38b2638df9c9e16a746c479452996b80d38L230

I mean `donation.name` will fall back to `sponsor.login` when `sponsor.name` is `null`. Considering that `login` can never be `null`, there is no need to set fallback.

chenxsan: Sure, will revert this change once review is done. Keeping it now so the CI can run without errors.

tools/fetch-sponsors.js
```diff
@@ -227,7 +227,7 @@ async function fetchGitHubSponsors() {
     // return an array in the same format as Open Collective
     const sponsors = organization.sponsorshipsAsMaintainer.nodes
         .map(({ sponsor, tier }) => ({
-            name: sponsor.name,
+            name: sponsor.name || sponsor.login,
```
harish-sethuraman, line 230: Is there anyway we'd not have both login and name values? 
I remember fixing the same for url once? https://github.com/eslint/eslint.org/pull/314/files

amareshsm: #314 this is for fixing URL only right?

harish-sethuraman: No I meant we have to fix the way I did if we thing sponsor.name or sponsor.login will be null? else this looks good to me

amareshsm: ya current PR itself looks good to me.

src/content/pages/sponsors.html
mdjermanovic: Can we update the data files in this PR instead?

tools/fetch-sponsors.js
mdjermanovic, line 230: I believe this fixes `null` names in `sponsors.json`, which is fine, but since we already had `null` names in that file (for example, [this one](https://github.com/eslint/eslint.org/blob/415e0bba84b8d205faa5c47924c928b0b0074dba/src/_data/sponsors.json#L1327-L1334)), build is probably failing because of `null` names in `donations.json` so it seems that we should do the same change few lines below:

https://github.com/eslint/eslint.org/blob/415e0bba84b8d205faa5c47924c928b0b0074dba/tools/fetch-sponsors.js#L242

src/content/pages/sponsors.html
mdjermanovic: This filter was broken because of multiple same (`null`) names:

https://github.com/eslint/eslint.org/blob/415e0bba84b8d205faa5c47924c928b0b0074dba/tools/fetch-sponsors.js#L249-L255

I think we can now remove this entry from `donations.json` as it's already in `sponsors.json` and that alone will probably fix the build:

https://github.com/eslint/eslint.org/blob/415e0bba84b8d205faa5c47924c928b0b0074dba/src/_data/donations.json#L254-L262

chenxsan: > Can we update the data files in this PR instead?

Absolutely! It's the best workaround actually.

tools/fetch-sponsors.js
chenxsan: > Is there anyway we'd not have both login and name values? I remember fixing the same for url once? https://github.com/eslint/eslint.org/pull/314/files

Nah, `login` is our github username, so it can't be `null`.

chenxsan:
Here's an update of the latest changes:

1. Use `sponsor.login` as the fallback value for `sponsor.name` in case it's `null`, there're two places in `fetch-sponsors.js`.
2. Clean up dirty data in `sponsors.json` to fix the CI, the latest data will be fetched at next midnight according to `data-fetch.yml` .

mdjermanovic:
I thought that we remove the entry from `donations.json` (as that's causing build failures) while keeping `sponsors.json` the same until the next automated fetch which will replace `null`s with usernames. Can you revert `sponsors.json`? If not, I'll merge as-is, since the entries will get back in a few hours anyways.

amareshsm (approved):
LGTM

amareshsm:
> I thought that we remove the entry from `donations.json` (as that's causing build failures) while keeping `sponsors.json` the same until the next automated fetch which will replace `null`s with usernames. Can you revert `sponsors.json`? If not, I'll merge as-is, since the entries will get back in a few hours anyways.

we can merge this.

mdjermanovic (approved):
LGTM, thanks!

PR #401, opened 2023-01-09
fix: adjust word order and enable animation on zh-hans

kecrily:
#### Prerequisites checklist

adjust word order and enable animation on zh-hans


src/_data/sites/zh-hans.yml
```diff
@@ -174,14 +174,14 @@ branding_page:
...
   
   animation:
-    enabled: false
-    text: 检测和修复 JavaScript 代码中的问题。
-    highlight_word: problems
-    wrong_word: problmes
+    enabled: true
+    text: 检测并修复 JavaScript 代码中的问题。
+    highlight_word: 代码中的问题。
+    wrong_word: 代码中的提问。
```
kecrily, line 184: The reason why Chinese and Japanese cannot use animate is that CJK text does not use the same space segmentation as Western text.

Here `animate.text` is word-segmented into `检测并修复`, `JavaScript` and `代码中的问题。`. But actually is `检测`, `并修复`, `JavaScript`, `代码中的` and `问题`. So I used this wrong participle to expand the scope of `highlight_word`.

It's not a very nice solution, but it works. Another problem now is that the lint content cannot be changed, and pictures are used.

截屏2023-01-09 23 01 45

aladdin-add: ideally the highlighted word should be "提问"(same as "problems"). just wondering if we can use an invisible zero-width char: "代码中的{zwc}问题"?

kecrily: I tried zero-width spaces, but the tokenizer doesn't recognize it.

aladdin-add: well, I think the current behaviour is fine.

aladdin-add (approved):
LGTM.

amareshsm (approved):
LGTM Thanks!

PR #421, opened 2023-03-03
chore: remove duplicate scss

harish-sethuraman:
Replacement of https://github.com/eslint/eslint.org/pull/365

Completely verified the changes locally and checked cascading of rules before and after removing duplicate selectors

### DIFF
--- a/src/assets/scss/forms.scss
+++ b/src/assets/scss/forms.scss
@@ -1,61 +1,3 @@
-.c-custom-select {
-    -moz-appearance: none;
-    -webkit-appearance: none;
-    appearance: none;
-    box-sizing: border-box;
-    display: block;
-    width: 100%;
-    max-width: 100%;
-    min-width: 0;
-    padding: 0.625rem 0.875rem;
-    padding-right: calc(0.875rem * 3);
-    padding-inline-end: calc(0.875rem * 3);
-    font: inherit;
-    color: var(--body-text-color);
-    color: inherit;
-    line-height: 1.3;
-    border: 1px solid var(--border-color);
-    border-radius: var(--border-radius);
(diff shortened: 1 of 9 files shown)


src/assets/scss/forms.scss
```diff
@@ -361,10 +301,6 @@ ul.selected-options {
     outline-offset: 3px;
 }
 
-.c-checkbox input[type="checkbox"]:focus ~ .c-checkbox__icon {
-    outline: 3px solid transparent;
-}
```
kecrily, line 366: There is a difference with the reserved line 360 line, which is 2px and here is 3px, but it is not a big problem.

src/assets/scss/forms.scss
```diff
@@ -1,61 +1,3 @@
...
-    background-size: 1em auto, 100%;
-}
-
-.label__text.label__text {
-    display: flex;
-    align-items: center;
-    gap: 0.5rem;
-    font-size: 0.875rem;
-    font-family: var(--text-font);
-    color: inherit;
```
kecrily, line 39: There is a difference from the reserved line 121, which is `color: var(--color-neutral-700);`, but I think the reserved one is the expected behavior.

kecrily (approved):
LGBT

src/assets/scss/forms.scss
harish-sethuraman: Seemed like 2px was being applied the whole time so removed the 3px one

src/assets/scss/forms.scss
harish-sethuraman: Yea this was getting over written by the above one so removed it

PR #454, opened 2023-07-25
fix: Exclude private sponsors from list

nzakas:
#### Prerequisites checklist

A few months ago we noticed that private sponsorships were being shown on the website. At the time, GitHub didn't have a way to filter out private sponsorships and also didn't provide any way to tell if a sponsorship was private, so we couldn't do anything about it.

I just received word that they have added in the ability to filter out private sponsorships and so I updated `fetch-sponsors.js` to do that.


amareshsm (approved):
LGTM. Leaving this open for anyone else to review before merging this.

Issue #423, opened 2023-03-08
Allow the selection of "all rules" in the playground


not-my-profile:
### What problem do you want to solve?

The [playground](https://eslint.org/play/) currently apparently only lets you select specific rules and does not provide a way to select all rules (which would be useful to find out whether or not eslint has a rule for a specific piece of code).

`eslint:all` has been implemented in https://github.com/eslint/eslint/commit/f804397f1a8d809c454d739b4adeb0a44e58600e but that option does not appear to be surfaced by the playground.

### What do you think is the correct solution?

Provide a way to select all rules in the playground.

### Participation

- [ ] I am willing to submit a pull request for this change.

not-my-profile:
Actually I think it might make more sense to simply allow the provision of an arbitrary `eslintrc` configuration, like the [typescript-eslint playground](https://typescript-eslint.io/play/) supports.

nzakas:
I'll leave this for @eslint/website-team to comment on, if anyone has any ideas.

At this point, we can't just allow any arbitrary config file to be uploaded because we are in the middle of changing the config system.

harish-sethuraman:
There are some things to addressed here. 
- If we consider allowing the eslint:all option to select all rules, Should we include it inside the dropdown? 
If we do so we can't maintain a state where rules can be opted out if not wanted
- If we show the use all rules option outside the dropdown as a checkbox we can show intermediate state as well. But would like to know how it would look in the UI

@kecrily @amareshsm Thoughts?

harish-sethuraman:
@eslint/website-team following up on this as it has been open for a long time?

amareshsm:
I missed this thread. Thanks for the reminder. 

FMPOV: I am ok with giving a checkbox to enable all the rules. Clicking on  `enable all the rules` (checked) option will add all the rules, and clicking again (unchecked) will remove all the rules.

And also we can give an option to reset to defaults. By clicking on reset to defaults it will reset all the configs to the default state. 

![image](https://user-images.githubusercontent.com/30730124/230687535-463a4beb-195d-4c3d-ad72-3d76e7161bfd.png)


PS: I will pick this one if we are implementing it.

kecrily:
I like the checkbox idea, but we should need an intermediate state to indicate that part of the rule is checked.

amareshsm:
> @amareshsm are you picking this? we have decided to settle up with keeping the checkbox that selects all the rules and delselects all the rules when un-checked.

[...]

PR #462, opened 2023-07-30
feat: Add "Enable all rules" option in playground

amareshsm:
#### Prerequisites checklist

Added "Enable all rules" option to enable all the rules in one go.

image

https://github.com/eslint/eslint.org/assets/30730124/6d7718e8-a23a-497c-9d81-8352b649c4b3

#### Related Issues
https://github.com/eslint/eslint.org/issues/423


nzakas:
I'm not sure the UI is right on this. If I check the checkbox then it adds all rules, but when I uncheck the box, I'm not sure what it's supposed to do.

amareshsm:
> I don't think a checkbox is the right way to implement this. Maybe we should use a button instead? So it says "Enable All Rules" and when you click on it, it changes to "Reset Rules" (or something similar)?

Yeah, we can have a button and change it accordingly. I will make the changes.

harish-sethuraman:
changes made by me:
-  changed the checkbox to be a button
- The select will also disappear when all the rules are selected. I did this because the select wont have anything when all rules are selected.

harish-sethuraman:
@nzakas can you take a look at this now? @eslint/website-team would like some review on this one?

src/assets/scss/components/playground-configuration.scss
```diff
@@ -136,6 +136,15 @@
...
 }
 
+.playground__config-enable_all_rules {
+    display: flex;
+    margin: 0.5rem 0;
+}
+
+.enable_all_rules-checkbox {
+    margin: 0 0.4rem 0.2rem 0;
+}
```
kecrily, line 146:
```suggestion
```

Now we're not using this class anymore.

harish-sethuraman: Removed it thanks. Just to confirm pushed a separate commit instead of committing suggestion

amareshsm:
LGTM, I have updated the PR description with the latest changes. Thanks @harish-sethuraman

nzakas:
@harish-sethuraman I'm still seeing the select even when all rules are enabled.
Screenshot 2023-12-28 at 14-00-44 ESLint Playground - ESLint - Pluggable JavaScript Linter

I also feel that the button is just way too big. I don't think this is a common use case, so it shouldn't be reflected with such a large UI element. Instead, I'd suggest adding a link next to "Add Rule" that says "Add All Rules".

harish-sethuraman:
@nzakas probably all the rules are not selected in the screenshot you shared? Can you try opening the dropdown to check if it is empty? because this is how it looks for me when all rules are selected. 

Screenshot 2023-12-29 at 12 58 56 PM

>  I'd suggest adding a link next to "Add Rule" that says "Add All Rules".

makes sense will change

nzakas:
I just checked again. When I click the button, I get errors in the console and nothing happens. I'm using Firefox.

I still think the button is too big regardless of the functionality.

harish-sethuraman:
Have fixed the above issue. and changed the button to a smaller one similar to add rule button 👍🏻 

https://deploy-preview-462--new-eslint.netlify.app/play/

nzakas:
Thanks! This is looking good. Only issue now: When I click "Enable all rules", it sets focus to the most recently added textbox. That works okay when there are no rules enabled, but if you already have a bunch of rules enabled, then it scrolls the entire side panel down. Can we change it so that it always sets focus to the first visible textbox instead?

amareshsm:
> > when we enable all rules it adds all rules and then focuses on the first one
> > ```
> > * when we already have some rules enabled and enable all rules again it focuses on the recently added one (alphabetical order)
> > ```
> 
> Yes, I'm saying this behavior is not desirable because it can cause the sidebar to scroll, which is very disorienting.

I have updated the PR accordingly now it always sets focus to the first visible textbox.

nzakas (approved):
Perfect! Thanks for all the work on this.

PR #485, opened 2023-10-25
chore: Remove unnecessary dependencies

nzakas:
#### Prerequisites checklist

Removed all of the metascraper-related dependencies as this appears to be the cause of CI failures like [this](https://github.com/eslint/eslint.org/actions/runs/6642763403/job/18058011707). The `metascraper-description` packages depends on `re2`, which is failing to compile on Node.js v20.8.1. This is likely because the dependency is out of date.

But since we aren't actually using metascraper here, I just removed it.


mdjermanovic (approved):
LGTM, thanks!

PR #506, opened 2024-01-02
chore: Update robots.txt to disallow crawling of non-latest docs

nzakas:
#### Prerequisites checklist

To ensure that only the latest version docs are crawled, disallowed crawling of `/docs/next/` and `/docs/head/`.

I also removed the old disallow when we were testing `new.eslint.org`.

### DIFF
--- a/src/static/robots.njk
+++ b/src/static/robots.njk
@@ -6,4 +6,5 @@ eleventyExcludeFromCollections: true
 Sitemap: https://{{ site.hostname }}/sitemap-index.xml
 
 User-agent: *
-{% if "new." in site.hostname %}Disallow: /{% endif %}
+Disallow: /docs/next/
+Disallow: /docs/head/


mdjermanovic (commented):
I'm not sure if this is the right approach. We already have `<meta name="robots" content="noindex">` on those pages, and some articles say that one should not be used with the other.

https://developers.google.com/search/docs/crawling-indexing/block-indexing

> Important: For the noindex rule to be effective, the page or resource must not be blocked by a robots.txt file, and it has to be otherwise accessible to the crawler. If the page is blocked by a robots.txt file or the crawler can't access the page, the crawler will never see the noindex rule, and the page can still appear in search results, for example if other pages link to it.

Might be good to ask for a SEO advice?

nzakas:
Ah, totally forgot about the meta tag. Nevermind. :)

PR #509, opened 2024-01-20
feat: migrate playground to flat config

snitin315:
#### Prerequisites checklist

Migration of the online playground to use flat config system.

- Update playground to use flat config.
- Removed environments.
- Supports existing links to the Playground by translating configurations to the new format.
- Updated download configuration functionality
   - ~If `sourceType` is `module` then the downloaded config is in the ESM format otherwise it's in the CommonJS format.~
   - Added an option to select config file format.
 
Screenshot 2024-01-26 at 4 38 58 PM

#### Related Issues

Fix https://github.com/eslint/eslint.org/issues/507

Should we update the default value for `sourceType` to `module`?

### DIFF
--- a/src/playground/App.js
+++ b/src/playground/App.js
@@ -6,41 +6,45 @@ import CrashAlert from "./components/CrashAlert";
 import Footer from "./components/Footer";
 import CodeEditor from "./components/CodeEditor";
 import { Linter, SourceCodeFixer } from "./node_modules/eslint/lib/linter/";
-import { Legacy } from "@eslint/eslintrc/universal";
 import Unicode from "./utils/unicode";
 import Configuration from "./components/Configuration";
 import Split from "react-split";
 import debounce from "./utils/debounce";
 import "./scss/split-pane.scss";
 
-const linter = new Linter();
-const rules = linter.getRules();
+const linter = new Linter({ configType: "flat" });
+const legacyLinter = new Linter();
+const rules = legacyLinter.getRules();
+
 const ruleNames = Array.from(rules.keys());
 const rulesMeta = Array.from(rules.entries()).reduce((result, [key, value]) => {
     result[key] = value.meta;
     return result;
 }, {});
 
-const defaultParserOptions = {
-    ecmaFeatures: {},
+const defaultLanguageOptions = {
     ecmaVersion: "latest",
-    sourceType: "script"
+    parserOptions: {
+        ecmaFeatures: {}
+    }
 };
 
 const fillOptionsDefaults = options =>
     (options
         ? {
-            env: {},
             rules: {},
             ...options,
-            parserOptions: {
-                ...defaultParserOptions,
-                ...options.parserOptions
+            languageOptions: {
+                ecmaVersion: "latest",
+                parserOptions: {
+                    ecmaFeatures: {},
+                    ...options.languageOptions?.parserOptions
+                },
+                ...options.languageOptions
             }
         }
         : {
-            env: { es6: true },
-            parserOptions: defaultParserOptions,
+            languageOptions: defaultLanguageOptions,
             rules: [...rules.entries()].reduce((result, [ruleId, rule]) => {
                 if (rule.meta.docs.recommended) {
                     result[ruleId] = ["error"];
@@ -49,9 +53,32 @@ const fillOptionsDefaults = options =>
             }, {})
         });
 
+const convertLegacyOptionsToFlatConfig = options => {
+    const { parserOptions } = options;
+    const flatConfigOptions = {
+        languageOptions: {
+            ecmaVersion: parserOptions.ecmaVersion || "latest",
+            parserOptions: {
+                ecmaFeatures: parserOptions.ecmaFeatures || {}
+            },
+            sourceType: parserOptions.sourceType || "script"
+        },
+        rules: options.rules
+
+    };
+
+    return flatConfigOptions;
+};
+
 const getUrlState = () => {
     try {
-        return JSON.parse(Unicode.decodeFromBase64(window.location.hash.replace(/^#/u, "")));
+        const urlOptions = JSON.parse(Unicode.decodeFromBase64(window.location.hash.replace(/^#/u, "")));
+
+        if (typeof urlOptions.options.languageOptions !== "undefined") {
+            return urlOptions;
+        }
+
+        return { text: urlOptions.text, options: convertLegacyOptionsToFlatConfig(urlOptions.options) };
     } catch {
         return {};
     }
@@ -72,20 +99,16 @@ const App = () => {
     const { text: urlText, options: urlOptions } = getUrlState();
     const [text, setText] = useState(urlText || storedText || "/* eslint quotes: [\"error\", \"double\"] */\nconst a = 'b';");
     const [fix, setFix] = useState(false);
-    const [options, setOptions] = useState(fillOptionsDefaults(urlText ? urlOptions || {} : storedOptions));
+
+    let flatConfigStoredOptions = storedOptions;
+
+    if (typeof storedOptions.languageOptions === "undefined") {
+        flatConfigStoredOptions = convertLegacyOptionsToFlatConfig(storedOptions);
+    }
+
+    const [options, setOptions] = useState(fillOptionsDefaults(urlText ? urlOptions || {} : flatConfigStoredOptions));
 
     const lint = () => {
-        try {
-            const validator = new Legacy.ConfigValidator({ builtInRules: rules });
-
-            validator.validate(options);
-        } catch (error) {
-            return {
-                messages: [],
-                output: text,
-                validationError: error
-            };
-        }
         try {
             const { messages, output } = linter.verifyAndFix(text, options, { fix });
             let fatalMessage;
@@ -203,6 +226,7 @@ const App = () => {
                         <CodeEditor
                             tabIndex="0"
                             codeValue={text}
+                            eslintInstance={linter}
                             eslintOptions={options}
                             onUpdate={debouncedOnUpdate}
                         />
--- a/src/playground/App.js
+++ b/src/playground/App.js
@@ -136,9 +136,23 @@ const App = () => {
     const [fix, setFix] = useState(false);
     const [options, setOptions] = useState(initialOptions);
 
+    // In some cases, Linter modifies `languageOptions`, so we'll deep-clone them
+    const optionsForLinter = {
+        ...options,
+        languageOptions: {
+            ...options.languageOptions,
+            parserOptions: {
+                ...options.languageOptions.parserOptions,
+                ecmaFeatures: {
+                    ...options.languageOptions.parserOptions.ecmaFeatures
+                }
+            }
+        }
+    };
+
     const lint = () => {
         try {
-            const { messages, output } = linter.verifyAndFix(text, options, { fix });
+            const { messages, output } = linter.verifyAndFix(text, optionsForLinter, { fix });
             let fatalMessage;
 
             if (messages && messages.length > 0 && messages[0].fatal) {
@@ -263,7 +277,7 @@ const App = () => {
                             tabIndex="0"
                             codeValue={text}
                             eslintInstance={linter}
-                            eslintOptions={options}
+                            eslintOptions={optionsForLinter}
                             onUpdate={debouncedOnUpdate}
                         />
                     </main>
--- a/src/playground/App.js
+++ b/src/playground/App.js
@@ -53,7 +53,7 @@ const fillOptionsDefaults = options =>
             }, {})
         });
 
-const convertLegacyOptionsToFlatConfig = options => {
+const convertLegacyOptionsToFlatConfig = (options = {}) => {
 
     // If options object is empty, return it
     if (options && Object.keys(options).length === 0 && options.constructor === Object) {
--- a/src/playground/App.js
+++ b/src/playground/App.js
@@ -59,6 +59,18 @@ const convertLegacyOptionsToFlatConfig = options => {
 
         flatConfigOptions.languageOptions = restParserOptions;
 
+        /*
+         * Normalize ecmaVersion to support links with configs generated
+         * by the old demo (https://archive.eslint.org/demo)
+         */
+        if (
+            typeof parserOptions.ecmaVersion === "number" &&
+            parserOptions.ecmaVersion >= 6 &&
+            parserOptions.ecmaVersion < 2015
+        ) {
+            flatConfigOptions.languageOptions.ecmaVersion = parserOptions.ecmaVersion + 2009;
+        }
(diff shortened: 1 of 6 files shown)


src/playground/components/Configuration.js
```diff
@@ -133,6 +129,8 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
         );
     };
 
+    const configFileContent = `${options.languageOptions?.sourceType === "module" ? "export default" : "module.exports ="} ${JSON.stringify([options], null, 4)};`;
```
nzakas, line 132: I don't think this is a safe assumption to make, at least, not based on the feedback we received about wanting `eslint.config.cjs` and `eslint.config.mjs` support. It's probably better to offer the choice of download format.

nzakas (changes requested):
Thanks for doing this. Just left a few comments that I believe answer most of your questions.

src/playground/components/Configuration.js
snitin315: Updated, now users can choose the format.
Screenshot 2024-01-25 at 3 48 39 PM

src/playground/App.js
```diff
@@ -6,41 +6,46 @@ import CrashAlert from "./components/CrashAlert";
...
         ? {
-            env: {},
             rules: {},
             ...options,
-            parserOptions: {
-                ...defaultParserOptions,
-                ...options.parserOptions
+            languageOptions: {
+                ecmaVersion: "latest",
+                sourceType: "script",
```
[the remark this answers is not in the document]
snitin315: Yes, udpated.

src/playground/components/Configuration.js
kecrily: This option is too far away from the download button. Maybe we can make it closer to the download button and remove the text about formatting on the button. Turning it into two buttons would be nice.

snitin315: I've moved the selector closer to the button.

Screenshot 2024-01-26 at 4 38 58 PM

nzakas:
This is looking really good! It appears there's a merge conflict, though.

snitin315:
@nzakas I've fixed the conflicts.

nzakas (approved):
LGTM, thanks! Leaving open to give others the opportunity to review before merging.

src/playground/App.js
```diff
@@ -6,41 +6,42 @@ import CrashAlert from "./components/CrashAlert";
...
+                }
             }
         }
         : {
-            env: { es6: true },
-            parserOptions: defaultParserOptions,
+            languageOptions: defaultLanguageOptions,
             rules: [...rules.entries()].reduce((result, [ruleId, rule]) => {
                 if (rule.meta.docs.recommended) {
                     result[ruleId] = ["error"];
```
mdjermanovic, line 47: When I open https://deploy-preview-509--new-eslint.netlify.app/play/ with a clean cache, recommended rules are not enabled, so it looks like this branch never happens. Maybe because `convertLegacyOptionsToFlatConfig` always returns an object?

src/playground/App.js
```diff
@@ -49,9 +50,39 @@ const fillOptionsDefaults = options =>
...
+    const { parserOptions } = options;
+    const flatConfigOptions = {
+        languageOptions: {
+            ecmaVersion: parserOptions.ecmaVersion || "latest",
+            parserOptions: {
+                ecmaFeatures: parserOptions.ecmaFeatures || {}
+            },
+            sourceType: parserOptions.sourceType || "module"
+        },
+        rules: options.rules
```
mdjermanovic, line 74: Links from the docs site don't work. For example, this is with `#...` copied from the first example for the [`accessor-pairs`](https://eslint.org/docs/latest/rules/accessor-pairs#setwithoutget) rule:

https://deploy-preview-509--new-eslint.netlify.app/play/#eyJvcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsic291cmNlVHlwZSI6Im1vZHVsZSJ9fSwidGV4dCI6Ii8qZXNsaW50IGFjY2Vzc29yLXBhaXJzOiBcImVycm9yXCIqL1xuXG52YXIgbyA9IHtcbiAgICBzZXQgYSh2YWx1ZSkge1xuICAgICAgICB0aGlzLnZhbCA9IHZhbHVlO1xuICAgIH1cbn07XG5cblxudmFyIG8gPSB7ZDogMX07XG5PYmplY3QuZGVmaW5lUHJvcGVydHkobywgJ2MnLCB7XG4gICAgc2V0OiBmdW5jdGlvbih2YWx1ZSkge1xuICAgICAgICB0aGlzLnZhbCA9IHZhbHVlO1xuICAgIH1cbn0pOyJ9

```
playground.js:2 TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at rv (playground.js:2:2203255)
    at wo (playground.js:2:131194)
    at Cl (playground.js:2:188149)
    at bc (playground.js:2:177131)
    at Ec (playground.js:2:177059)
    at yc (playground.js:2:176922)
    at oc (playground.js:2:173729)
    at sc (playground.js:2:172284)
    at C (playground.js:2:279135)
```

When I try this locally (with the debug info), it seems that the error is thrown when trying to do `Object.keys(options.rules)`.

I _think_ this happens because this function always creates `rules` property, even if the value is `undefined`, so in `fillOptionsDefaults` it overwrites `rules: {}`.

snitin315: I think this broke after I rebased on top of https://github.com/eslint/eslint.org/commit/44cfc74d2de96c9bcaa6c143b1a372fc938abb5f, I'll push a fix.

src/playground/App.js
```diff
@@ -72,20 +103,9 @@ const App = () => {
...
-            const validator = new Legacy.ConfigValidator({ builtInRules: rules });
-
-            validator.validate(options);
-        } catch (error) {
-            return {
-                messages: [],
-                output: text,
-                validationError: error
-            };
-        }
```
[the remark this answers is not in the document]
snitin315: Good Catch! Ideally, it would be great if we could set an `error.code = "ESLINT_VALIDATION_ERROR` or a specific exit code in case of validation errors. Right now I'm evaluating it based on the presence of`key "rules"` in error message.

src/playground/App.js
snitin315: You're right. Pushed a fix.

snitin315:
@mdjermanovic I've resolved your comments

src/playground/App.js
```diff
@@ -6,41 +6,44 @@ import CrashAlert from "./components/CrashAlert";
...
             rules: {},
             ...options,
-            parserOptions: {
-                ...defaultParserOptions,
-                ...options.parserOptions
+            languageOptions: {
+                ...options.languageOptions,
+                parserOptions: {
+                    ecmaFeatures: {},
+                    ...options.languageOptions?.parserOptions
```
aladdin-add, line 41: it will throw an error when `options.languageOptions` is `null/undef`?

src/playground/components/Configuration.js
```diff
@@ -184,10 +183,12 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
                                 isSearchable={false}
                                 styles={customStyles}
                                 theme={theme => customTheme(theme)}
-                                defaultValue={ECMAVersionsOptions.filter(ecmaVersion => options.parserOptions.ecmaVersion === ecmaVersion.value)}
+                                defaultValue={
+                                    ECMAVersionsOptions.filter(ecmaVersion => ecmaVersion.value === (options.languageOptions.ecmaVersion || "latest"))
```
aladdin-add, line 187: Set to "latest" if it's invalid?
```suggestion
                                    ECMAVersionsOptions.filter(ecmaVersion => ecmaVersion.value === (options.languageOptions.ecmaVersion)) || "latest"
```

src/playground/App.js
```diff
@@ -49,9 +52,39 @@ const fillOptionsDefaults = options =>
...
 
+const convertLegacyOptionsToFlatConfig = (options = {}) => {
+
+    // If options object is empty, return it
+    if (isEmptyObject(options)) {
+        return options;
+    }
+
+    // If options is already a flat config, return it
+    if (typeof options.languageOptions !== "undefined" && typeof options.parserOptions === "undefined") {
```
aladdin-add, line 63: should also check `env` - it's away in the new config.

src/playground/components/Configuration.js
```diff
@@ -199,10 +200,10 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
                             isSearchable={false}
                             styles={customStyles}
                             theme={theme => customTheme(theme)}
-                            defaultValue={sourceTypeOptions.filter(sourceTypeOption => options.parserOptions.sourceType === sourceTypeOption.value)}
+                            defaultValue={sourceTypeOptions.filter(sourceTypeOption => sourceTypeOption.value === (options.languageOptions.sourceType || "module"))}
```
aladdin-add, line 203: same here?

src/playground/components/Configuration.js
```diff
@@ -142,7 +140,7 @@ export default function Configuration({ rulesMeta, eslintVersion, onUpdate, opti
         );
     };
 
-    const allRulesSelected = Object.keys(options.rules).length === ruleNames.length;
+    const allRulesSelected = Object.keys(options.rules || {}).length === ruleNames.length;
```
mdjermanovic, line 143: Same question here.

nzakas:
@snitin315 are you still working on this?

mdjermanovic:
Since @snitin315 is away, I'll take over this branch to finish this up.

src/playground/App.js
mdjermanovic: `{ ...undefined }` doesn't throw, it's just a no-op.

mdjermanovic:
> > If we want to rely on Linter's defaults, then I think the dropdowns should have an empty item that would be selected by default. Otherwise, we could set defaults in the Playground's config as it used to be before this change, which I think is fine.
> 
> Rather than an empty item, maybe "(default)" for clarity? I'd prefer not to always include `sourceType` and `ecmaVersion` in the config, as we should be encouraging people to use the defaults unless they have a specific need.

I've added a "(default)" item in the two lists.

src/playground/App.js
mdjermanovic: This is now refactored a bit, please check the new version.

src/playground/components/Configuration.js
mdjermanovic: https://github.com/eslint/eslint.org/pull/509#discussion_r1541507365

mdjermanovic:
Ready for re-review.

nzakas:
I tested a few different combinations and verified that the `ecmaVersion` and `sourceType` fields do not appear in config files when the default values are used.

I did notice that `parserOptions.ecmaFeatures` is always present, even when I didn't add any ecmaFeatures. Is that expected?

mdjermanovic:
> I did notice that `parserOptions.ecmaFeatures` is always present, even when I didn't add any ecmaFeatures. Is that expected?

It works that way in the current https://eslint.org/play/, and in the old https://archive.eslint.org/demo too.

I'm not sure if this was intentional. It could be that the idea was to always create an object, even if empty, so that the user could easily add to it later.

Either way, I think it's fine to remove `ecmaFeatures` and its parent `parserOptions` object altogether if no ECMA Features were selected.

Furthermore, we could also remove `languageOptions` if there are no ECMA Features selected and ECMA Version & Souce Type are both "(default)"?

src/playground/App.js
```diff
@@ -6,54 +6,92 @@ import CrashAlert from "./components/CrashAlert";
...
 import debounce from "./utils/debounce";
 import "./scss/split-pane.scss";
 
-const linter = new Linter();
-const rules = linter.getRules();
+const DEFAULT_TEXT = "/* eslint quotes: [\"error\", \"double\"] */\nconst a = 'b';";
+
+const linter = new Linter({ configType: "flat" });
+const legacyLinter = new Linter({ configType: "eslintrc" });
+const rules = legacyLinter.getRules();
```
aladdin-add, line 19: we had reached a consensus to https://github.com/eslint/eslint/issues/18103#issuecomment-1941230245.

mdjermanovic: Agreed, but that will need some time to implement, so I think we can leave it for a follow-up. This will work until at least v10 :)

aladdin-add: haha, it also makes sense.

aladdin-add (approved):
LGTM, pending @nzakas 's review.

nzakas:
> > I did notice that `parserOptions.ecmaFeatures` is always present, even when I didn't add any ecmaFeatures. Is that expected?
> 
> It works that way in the current https://eslint.org/play/, and in the old https://archive.eslint.org/demo too.
> 
> I'm not sure if this was intentional. It could be that the idea was to always create an object, even if empty, so that the user could easily add to it later.
> 
> Either way, I think it's fine to remove `ecmaFeatures` and its parent `parserOptions` object altogether if no ECMA Features were selected.
> 
> Furthermore, we could also remove `languageOptions` if there are no ECMA Features selected and ECMA Version & Source Type are both "(default)"?

Yes, I think we don't need any object that doesn't have any overrides in it.

mdjermanovic:
> > > I did notice that `parserOptions.ecmaFeatures` is always present, even when I didn't add any ecmaFeatures. Is that expected?
> > 
> > 
> > It works that way in the current https://eslint.org/play/, and in the old https://archive.eslint.org/demo too.
> > I'm not sure if this was intentional. It could be that the idea was to always create an object, even if empty, so that the user could easily add to it later.
> > Either way, I think it's fine to remove `ecmaFeatures` and its parent `parserOptions` object altogether if no ECMA Features were selected.
> > Furthermore, we could also remove `languageOptions` if there are no ECMA Features selected and ECMA Version & Source Type are both "(default)"?
> 
> Yes, I think we don't need any object that doesn't have any overrides in it.

Done in https://github.com/eslint/eslint.org/pull/509/commits/ae67deb7755ec279cdb4578394646b9d48e21df8.

I noticed that in some cases Linter modifies passed configs (for example, [here](https://github.com/eslint/eslint/blob/9b7bd3be066ac1f72fa35c4d31a1b178c7e2b683/lib/linter/linter.js#L1686)), which causes confusing behavior in the Playground, so I've updated code to pass a cloned config in https://github.com/eslint/eslint.org/pull/509/commits/f5ecbe7fa30d4bde313af4ee3a46568ca3f59435.

I also added ecmaVersion normalization in https://github.com/eslint/eslint.org/pull/509/commits/e4ae6cd7b264ab83241682cd2a58a03b36374616, as per the discussion in https://github.com/eslint/eslint/pull/18241.

nzakas (approved):
LGTM. Thanks!

PR #578, opened 2024-06-11
chore: Add flex-wrap property for responsive layout on homepage

smadixd:
#### Prerequisites checklist

Handle responsive layout of metrics in home page on ultra wide monitors/views.

Basically adding a flex wrap property to handle in the case of step-6 font being too large on ultra wide views which causes the text to get out of its parent frames.

here's a before and after screenshots

ultra wide:
image
image

wide:
image
image

normal:
image
image

extra small/mobile:
image
image

### DIFF
--- a/src/assets/scss/foundations.scss
+++ b/src/assets/scss/foundations.scss
@@ -114,6 +114,10 @@ main {
     max-width: 1280px;
     margin: 0 auto;
     padding: var(--space-xl-3xl) calc(1rem + 1vw);
+
+    @media screen and (min-width: 1920px) {
+        max-width: 60%;
+    }
 }
 
 .section-head {
--- a/src/assets/scss/homepage.scss
+++ b/src/assets/scss/homepage.scss
@@ -173,7 +173,6 @@
     @media all and (min-width: 768px) {
         display: flex;
         justify-content: space-evenly;
-        flex-wrap: wrap;
     }
 }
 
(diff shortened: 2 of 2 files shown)


nzakas:
Thanks for the PR. What dimensions are you considering "wide" and "ultra wide"? That would be helpful to verify this change.

smadixd:
Hi @nzakas 

I'm using an ultrawide LG monitor however now looking at the dimensions 

the full screen resolution (ultra wide) was at  **3440px x 1270px**
3/4 of the screen resolution is what i considered (wide) was at **2440px x 1270px**

nzakas:
Thanks, I was able to verify that this was happening in Chromium but not in Firefox.

I'm not sure I understand the numbers need to wrap at such a large resolution, though. There's plenty of horizontal space. Is there something we can do to make all of the numbers stay on the same line?

smadixd:
@nzakas this is because of the step-6 font size causing the font to become too large at ultra wide views, changing it or its calculation might not be ideal as it might be used in other places, however we can set a smaller font size on large views. using media queries or even setting the font based on view width.

smadixd:
@nzakas The issue was with the max-width property on the `content-container` class. it persisted the width of the content even on large views, I have pushed new changes now, this adds a media query check on larger screens to change the max-width of the containers to use the extra space on the sides.

nzakas (approved):
LGTM. Thank you!

Issue #587, opened 2024-06-22
Suggestion: [playground] Error Row:Col to be Clickable to Navigate and Focus on Specific Code Location


smadixd:
### What problem do you want to solve?

Basically make the line:column clickable and scroll to that line and move cursor to column as shown in screenshot below

![image](https://github.com/eslint/eslint.org/assets/49820024/e9c450cf-683d-4e9e-a6c5-bbbf49e87aa2)

### What do you think is the correct solution?

I haven't looked at the playground code yet but I'd like to take a look on this.

### Participation

- [X] I am willing to submit a pull request for this change.

amareshsm:
I'm fine with this change

snitin315:
Feel free to send a PR.

PR #590, opened 2024-06-23
feat: [playground]Add scrollToPosition functionality to Editor component

smadixd:
#### Prerequisites checklist

This PR handles adding functionality to the playground where the user would have the ability to click on a row:col of an alert and would focus and redirection selection/cursor to that position.

Changes to the playground code editor component to inject an extra functionality to scroll to a position.

#### Related Issues

#587

### DIFF
--- a/src/playground/components/CodeEditor.js
+++ b/src/playground/components/CodeEditor.js
@@ -1,4 +1,4 @@
-import React, { useMemo } from "react";
+import React, { forwardRef, useImperativeHandle, useMemo, useRef } from "react";
 import CodeMirror from "@uiw/react-codemirror";
 import { history } from "@codemirror/history";
 import { bracketMatching } from "@codemirror/matchbrackets";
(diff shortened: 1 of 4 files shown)


src/playground/components/CodeEditor.js
```diff
@@ -10,33 +10,72 @@ import {
...
+			[eslintOptions, eslintInstance],
+		);
+
+		useImperativeHandle(ref, () => ({
+			scrollToPosition(line, col) {
+				if (editorRef.current) {
+					const editorView = editorRef.current.view;
+					if (editorView) {
+						const { state } = editorView;
+						const pos = state.doc.line(line).from + col;
```
mdjermanovic, line 35: Column where the cursor is after clicking the 2:11 link seems to be off-by-one:

image

Note that in ESLint messages both lines and columns are 1-based.

smadixd: so in this case it should be after the b character?

mdjermanovic: The other way, it should be before the left quote.

src/playground/components/CodeEditor.js
```diff
@@ -10,33 +10,72 @@ import {
...
+				bracketMatching(),
+				linter(esLint(eslintInstance, eslintOptions), { delay: 0 }),
+				javascript(),
+				ESLintPlaygroundTheme,
+				ESLintPlaygroundHighlightStyle,
+			],
+			[eslintOptions, eslintInstance],
+		);
+
+		useImperativeHandle(ref, () => ({
```
amareshsm, line 29: Is it necessary to use ImperativeHandle for this use case? 

Instead, we could store the line and column in a state variable and pass it as a prop to the code editor. In the code editor component, we can use a regular ref to get the editor view and then use the passed props for the line and column to point to the specific line. Are there any drawbacks to this approach?

smadixd: Using `useImperativeHandle` allows the parent component to directly control the editor, ensuring precise and immediate scrolling and focusing. Using `state` and `useEffect` can lead to issues with redundant state updates and synchronization, such as when clicking the same row and column twice. This approach keeps the scrolling logic encapsulated in the `CodeEditor`.

amareshsm: cool

mdjermanovic (dismissed):
LGTM, thanks! Leaving open for @amareshsm to verify, and would also like @snitin315 to verify.

src/playground/components/CodeEditor.js
```diff
@@ -10,33 +10,72 @@ import {
...
+							const offset =
+								linePos.top -
+								editorRect.top -
+								editorRect.height / 2;
+							editorView.scrollDOM.scrollTop += offset;
+						}
+
+						editorView.focus();
+					}
+				}
```
amareshsm, line 60:
```suggestion
                const editorView = editorRef.current?.view;
                if (editorView) {
                    const { state } = editorView;
                    const pos = state.doc.line(line).from + col;

                    // Set the cursor selection to the position
                    editorView.dispatch({
                        selection: { anchor: pos },
                        scrollIntoView: true,
                    });

                    const linePos = editorView.coordsAtPos(
                        state.doc.line(line).from,
                    );

                    // Calculate to try to center the line in the editor
                    if (linePos) {
                        const editorRect =
                            editorView.dom.getBoundingClientRect();
                        const offset =
                            linePos.top -
                            editorRect.top -
                            editorRect.height / 2;
                        editorView.scrollDOM.scrollTop += offset;
                    }

                    editorView.focus();
                }
```

Removed Unnecessary Null Checks for editorRef.current and editorRef.current.view

amareshsm (approved):
LGTM, Thanks. Waiting for Nithin's review.

snitin315 (approved):
LGTM, thanks!

PR #609, opened 2024-08-01
feat: add EOF banner

amareshsm:
#### Prerequisites checklist

Added EOL banner. (Addressed point no. 1 https://github.com/eslint/eslint.org/issues/594#issuecomment-2243119666)

| Desktop | Mobile |
---------|---------
| Screenshot 2024-08-03 at 1 35 01 AM |  Screenshot 2024-08-03 at 1 34 33 AM |
| Screenshot 2024-08-03 at 1 35 15 AM |  Screenshot 2024-08-03 at 1 34 10 AM |

#### Related Issues
https://github.com/eslint/eslint.org/issues/594

~~1. banner colour option selection for both dark and light themes. 
2. which link should be associated with the "long-term support" option text?~~


src/assets/scss/homepage.scss
```diff
@@ -280,3 +280,44 @@
...
+	display: flex;
+	justify-content: center;
+	align-items: center;
+	padding: 2px 0;
+	background-color: var(--link-color);
+	color: var(--body-background-color);
+
+	a {
+		color: var(--body-background-color);
+		padding: 0 4px;
```
harish-sethuraman, line 295: I don't think you would need padding to anchor element if you were using `<p>` tag for content? It would preserve white space for you?

nzakas:
Thanks for putting this together. It looks like the hourglass is aligned to the text baseline, but it should be vertically centered in the banner. It appears there may be some extra space at the bottom of the hour glass?
Screenshot 2024-08-02 at 10-05-27 Find and fix problems in your JavaScript code - ESLint - Pluggable JavaScript Linter

We had decided not to use the bright purple banner as it was too aggressive. We were going to use the gray in option 1 of [this comment](https://github.com/eslint/eslint.org/issues/594#issuecomment-2211303256).

"Long-term support options" should link to /version-support.

I would also adjust the text in the banner to read:

> ESLint v8.x end-of-life is 2024-10-05 and will not be maintained after that. [Upgrade](/docs/latest) or consider [long-term support options](/version-support).

src/assets/scss/homepage.scss
amareshsm: Yes, we can use the `p` tag and remove the margin. However, in the mobile view, the text jumps to a new line because the p tag uses the `display: block` property.

src/_includes/layouts/base.html
```diff
@@ -155,6 +155,25 @@
...
+                    </path>
+                </g>
+            </g>
+        </svg>
+        ESLint v8.x end-of-life is 2024-10-05 and will not be maintained after that. <a href="https://eslint.org/docs/latest/"> Upgrade </a>
+        or consider <a href="https://eslint.org/version-support/"> long-term support options </a> </div>
```
nzakas, line 176: Let's make these relative to ensure we're navigating within the same site.
```suggestion
        ESLint v8.x end-of-life is 2024-10-05 and will not be maintained after that. <a href="/docs/latest/"> Upgrade </a>
        or consider <a href="/version-support/"> long-term support options </a> </div>
```

amareshsm: done

src/_includes/layouts/base.html
```diff
@@ -155,6 +155,25 @@
...
+                    </path>
+                </g>
+            </g>
+        </svg>
+        ESLint v8.x end-of-life is 2024-10-05 and will not be maintained after that. <a href="/docs/latest/"> Upgrade </a>
```
aladdin-add, line 175: is it better to update the link to migration guide:
```suggestion
        ESLint v8.x end-of-life is 2024-10-05 and will not be maintained after that. <a href="/docs/latest/use/migrate-to-9.0.0"> Upgrade </a>
```

nzakas: good suggestion!

PR #615, opened 2024-08-20
fix: make algolia link not reachable via tab key

chenxsan:
#### Prerequisites checklist

When searching with the search box, say I input `mjs` and press `<Tab>` key, it would focus firstly on the link of algolia which  is covered by the search result modal, then the clear button, then the result item. IMO it makes more sense to disable it from reachable via tab key.

Screenshot-r7HoVxar@2x

Add `tabindex="-1"` to `<a />` element.


harish-sethuraman (approved):
LGTM

amareshsm (approved):
LGTM

nzakas:
Hey folks - I don't think this is the right approach to solve this problem. All links should be accessible via tab. The only issue here is that we want tab to go into the dropdown and not to the link. There's got to be a better solution that preserves the tab accessibility of the link.

Issue #598, opened 2024-07-05
Suggestion: Add playground quick links to open a new issue


JoshuaKGoldberg:
### What problem do you want to solve?

Sometimes I open the playground to confirm a reproduction of a rule bug or feature request, then immediately want to file a new issue with the playground. But there's no one-click way to do that. We have to manually copy the environment, configuration, and Minimal Reproduction Example into templates like [🐞 Report a problem](https://github.com/eslint/eslint/issues/new?assignees=&labels=bug%2Crepro%3Aneeded&projects=&template=bug-report.yml&title=Bug%3A+%28fill+in%29). That's inconvenient, and users don't always fill out the templates correctly.

### What do you think is the correct solution?

Proposal: let's add a button to the playground with a name like _"Report an issue"_ that links to a pre-filled issue template?

For example, given https://github.com/eslint/eslint/issues/18606's https://eslint.org/play/#eyJ0ZXh0IjoidHJ5IHtcbiAgdGhyb3cgbmV3IEVycm9yKCk7XG59IGNhdGNoIChfKSB7IFxufSIsIm9wdGlvbnMiOnsicnVsZXMiOnsibm8tdW51c2VkLXZhcnMiOlsid2FybiIseyJjYXVnaHRFcnJvcnNJZ25vcmVQYXR0ZXJuIjoiZm9vIn1dfSwibGFuZ3VhZ2VPcHRpb25zIjp7InNvdXJjZVR5cGUiOiJtb2R1bGUiLCJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319fX19, it would link to the equivalent of:

```js
const url = new URL("https://github.com/eslint/eslint/issues/new?assignees=&labels=bug%2Crepro%3Aneeded&projects=&template=bug-report.yml");
url.searchParams.set("title", "Bug: [no-unused-vars] (fill in)");

// We'd need to add 'id's to the .github/bug-report.yml reproduction input
[...]

PR #616, opened 2024-08-20
feat: add report an issue button

harish-sethuraman:
#### Prerequisites checklist

- Added report an issue button

- added button that allows pre filling data while creating bug reports in `eslint` repo

#### Related Issues

fixes https://github.com/eslint/eslint.org/issues/598

- pre-filled the reproducible example url field
-  pre-filled output field

should we do it for any other field?

### DIFF
--- a/src/playground/components/ShareURL.js
+++ b/src/playground/components/ShareURL.js
@@ -1,6 +1,6 @@
 import React, { useState } from "react";
 
-export default function ShareURL({ url }) {
+export default function ShareURL({ url, errors }) {
 	const [isDataCopied, setIsDataCopied] = useState(false);
 	const [showShareURL, setShowShareURL] = useState(false);
 
@@ -10,7 +10,7 @@ export default function ShareURL({ url }) {
 				className="playground-toggle c-btn c-btn--ghost"
 				onClick={() => setShowShareURL(currentValue => !currentValue)}
 			>
-				<h2 data-config-section-title>Share URL</h2>
+				<h2 data-config-section-title>Share</h2>
 				<svg
 					height="20"
 					width="20"
@@ -89,6 +89,34 @@ export default function ShareURL({ url }) {
 							</span>
 						)}
 					</div>
+
+					<button
+						onClick={() => {
+							const reportUrl = new URL(
+								"https://github.com/eslint/eslint/issues/new",
+							);
+							const params = {
+								labels: "bug,repro:yes",
+								template: "bug-report.yml",
+								"repro-url": window.location,
+								"lint-output": errors
+									.map(
+										({ line, column, message, ruleId }) =>
+											`${line}:${column} ${message} (${ruleId})`,
+									)
+									.join("\n"),
+							};
+
+							Object.entries(params).forEach(([key, value]) => {
+								reportUrl.searchParams.append(key, value);
+							});
+
+							window.open(reportUrl, "_blank");
+						}}
+						className="c-btn c-btn--secondary report-issue-btn"
+					>
+						Report an issue
+					</button>
 				</div>
 			)}
 		</div>
--- a/src/assets/scss/components/playground-configuration.scss
+++ b/src/assets/scss/components/playground-configuration.scss
@@ -242,6 +242,11 @@
 	}
 }
 
+.report-issue-btn {
+	width: 100%;
+	margin-top: 20px;
+}
+
 .playground__config-options__section--rules {
 	font-size: 0.875rem;
 }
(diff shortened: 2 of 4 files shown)


src/playground/components/ShareURL.js
```diff
@@ -89,6 +89,34 @@ export default function ShareURL({ url }) {
...
+
+					<button
+						onClick={() => {
+							const reportUrl = new URL(
+								"https://github.com/eslint/eslint/issues/new",
+							);
+							const params = {
+								labels: "bug,repro:yes",
+								template: "bug-report.yml",
+								"repro-url": window.location,
```
amareshsm, line 101:
```suggestion
								"repro-url": window.location.href,
```
We only need the URL, not the entire `window.location` object I guess.

nzakas:
Is it possible to also include the configuration information in the issue?

amareshsm:
> Is it possible to also include the configuration information in the issue?



>> The playground link will have the configuration. Should we add the configuration inside `What did you do?` in case we want it apart from having it in playground @nzakas

+1


We should be able to get the necessary configuration information from the playground link, which should be sufficient I guess. If you'd like, we can also include the configuration details explicitly. We already have the configFileContent that we use for downloading the configuration, so we can use the same approach for this.

nzakas:
Yes, I think it would be nice to output the configuration into the issue as well. It just makes it easier when reading the issue.

harish-sethuraman:
To prefill the configuration to description box need to merge this https://github.com/eslint/eslint/pull/18817. This was a miss as last time I thought we wouldn't want to prefill description. Should we add ids to all elements?

harish-sethuraman:
We will now add configuration file's content to report `what did you do` section. When lot of rules are selected the content becomes pretty long and github breaks when we try sending pretty long code through params. That is why I haven't added formatting to the config.

nzakas:
Unfortunately, the configuration really isn't readable like this if there's more than a couple of rules. Is it not possible to add the three backticks before and after, or otherwise to format it a bit?

Can we include the code, too?

And can we include the playground link to jump right back to it?

harish-sethuraman:
> Unfortunately, the configuration really isn't readable like this if there's more than a couple of rules. Is it not possible to add the three backticks before and after, or otherwise to format it a bit?

If we add formatting to the rules and there are huge number of rules (seems like it breaks when I added more than 53 rules) then Im getting this while I redirect to github for creating a new issue. 


Screenshot 2024-08-28 at 8 40 52 PM


> Can we include the code, too?

Same for code if we start adding more code we would not be able to create a gh issue. Seems like its same with TSESLint's playground as well it breaks when we hit a limit :( 



> And can we include the playground link to jump right back to it?

we are adding the playground URL in the `Link to Minimal Reproducible Example` section

amareshsm:
Should we disable the button with some error message when the users add many rules (i.e. more than 53 rules)?

nzakas:
It might be nice to just check the length of the URL and show an error message in the playground when they click Share URL

harish-sethuraman:
Im not sure about the length that github supports. I can see that TSESLint playground also has the same issue where if we write way too much code we get the exact same error in github.

amareshsm:
> > Should we disable the button with some error message when the users add many rules (i.e. more than 53 rules)?
> 
> it also depends on how much code the user has written so we can't rely only on the number of rules the user has selected.
> 
> Im not sure about the length that github supports. I can see that TSESLint playground also has the same issue where if we write way too much code we get the exact same error in github.


Yes, we can't display an error message based on the number of rules, but we can validate the length of the URL and ensure it doesn't exceed a certain limit.

nzakas:
> Yes, we can't display an error message based on the number of rules, but we can validate the length of the URL and ensure it doesn't exceed a certain limit.

That seems like a good idea. Just pop up an error if the URL gets too long. I think that's a better experience than letting the browser error out.

mdjermanovic:
Prettier playground saves the issue body to the clipboard when it's too long:

image

harish-sethuraman:
The total supported length of the search param seems to be `8148` characters. I'll add a comment like @mdjermanovic has shared that says paste from clipboard. This will allow users to paste the config into the issue? is that fine?

harish-sethuraman:
fixed it

harish-sethuraman:
@eslint/eslint-tsc can we close this please?

harish-sethuraman:
1. The code causing the error doesn't show up in the issue - What to do if the config is already too big? we need to choose between showing the config / code? 
- The URL back to the Playground is the default link, so clicking on it doesn't give you a reproducible case - Its probably because maybe you did not modify the code? 🤔 
@nzakas 
Screenshot 2025-05-11 at 3 40 58 PM

nzakas:
Closed in favor of https://github.com/eslint/eslint.org/pull/724

PR #617, opened 2024-08-20
Revert "fix: make algolia link not reachable via tab key"

nzakas:
Links should be accessible via tab. We shouldn't eliminate that. We need a different approach to solve the original problem.

Reverts eslint/eslint.org#615


harish-sethuraman:
Screenshot 2024-08-20 at 8 01 58 PM
These links are still accessible by tabs?

amareshsm (approved):
LGTM

Issue #653, opened 2024-10-24
Suggestion: Include thanks.dev sponsors


nzakas:
### What problem do you want to solve?

Right now, we are getting sponsorship money from thanks.dev, but we aren't displaying any of those sponsors on the website.

### What do you think is the correct solution?

Update the `fetch-sponsors.js` script to include thanks.dev sponsors.

Here's the API:
https://api.thanks.dev/v1/entity/gh/eslint/donors

For each donor, we should look at the last item in the `payments` array and, if the month is the current month, include them as a sponsor. For example, we are currently in October 2024, so only payments marked as `2024-10` should be included.

### Participation

- [ ] I am willing to submit a pull request for this change.

harish-sethuraman:
Ill pick this

nzakas:
Thanks!

PR #655, opened 2024-10-25
chore: add thanks.dev

harish-sethuraman:
#### Prerequisites checklist

closes https://github.com/eslint/eslint.org/issues/653

- added thanks.dev donors inside sponsors 
- modified the return type. Its not an array instead its a promise with object and its values

#### Related Issues
- https://github.com/eslint/eslint.org/issues/653

- just confirm if we want this to be inside sponsors or donations

### DIFF
--- a/tools/fetch-sponsors.js
+++ b/tools/fetch-sponsors.js
@@ -100,7 +100,9 @@ function getTierSlug(monthlyDonation) {
 
 /**
  * Fetch order data from Open Collective using the GraphQL API.
- * @returns {Array} An array of sponsors.
+ * @returns {Promise<{sponsors: Array, donations: Array}>} An promise that resolves to an object with two properties:
+ *  - `sponsors` (Array): List of sponsors
+ * - `donations` (Array): List of donations
  */
 async function fetchOpenCollectiveData() {
 	const endpoint = "https://api.opencollective.com/graphql/v2";
@@ -205,7 +207,9 @@ async function fetchOpenCollectiveData() {
 
 /**
  * Fetches GitHub Sponsors data using the GraphQL API.
- * @returns {Array} An array of sponsors.
+ * @returns {Promise<{sponsors: Array, donations: Array}>} An promise that resolves to an object with two properties:
+ *   - `sponsors` (Array): List of sponsors
+ *  - `donations` (Array): List of donations
  */
 async function fetchGitHubSponsors() {
 	function sponsorshipsQuery(cursor = null) {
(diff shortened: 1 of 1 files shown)


amareshsm:
> * just confirm if we want this to be inside sponsors or donations

Since we can't distinguish between donations and sponsors from the data in the thanks.dev API, it seems we should treat these as donations.

tools/fetch-sponsors.js
```diff
@@ -402,6 +457,7 @@ async function fetchGitHubSponsors() {
 	] = await Promise.all([
 		fetchOpenCollectiveData(),
 		fetchGitHubSponsors(),
+		fetchThanksDevData(),
```
amareshsm, line 460:
```suggestion
		fetchThanksDevData(),
```
The result of fetchThanksDevData() is not assigned to any variable or used in the destructuring. not sure if you are waiting for the confirmation.

harish-sethuraman:
> Since we can't distinguish between donations and sponsors from the data in the thanks.dev API, it seems we should treat these as donations.

the issue says we need to add the users to sponsors. Seems like I missed to push the latter commit will push.

tools/fetch-sponsors.js
```diff
@@ -387,6 +391,57 @@ async function fetchGitHubSponsors() {
...
+
+	if (process.env.DEBUG) {
+		await fs.writeFile(
+			"./debug-thanksdev-raw-response.json",
+			JSON.stringify(payload, null, 4),
+			{ encoding: "utf8" },
+		);
+	}
+
+	const sponsors = Object.values(payload.donors).reduce(
```
nzakas, line 417: I think this may be clearer by doing a `filter()` and then a `map()`. First filter out all of the donors that don't have a donation in the current month, then map into the final form.

`reduce()` is really for calculating a single value based on an array of values.

tools/fetch-sponsors.js
```diff
@@ -387,6 +391,57 @@ async function fetchGitHubSponsors() {
...
+		(accumulator, { name, login, avatar, url, payments }) => {
+			const lastPayment = payments.at(-1);
+
+			const paymentMonth = new Date(lastPayment.month).getMonth();
+			const currentMonth = new Date().getMonth();
```
nzakas, line 422: We also need to compare the year to make sure it's the current year.

nzakas (commented):
Thanks for taking a look at this. I left a couple notes on the algorithm.

And yes, we want these folks to end up in the sponsors list.

tools/fetch-sponsors.js
harish-sethuraman: updated

tools/fetch-sponsors.js
```diff
@@ -45,7 +45,7 @@ const knownOneTimers = new Set([
 const { ESLINT_GITHUB_TOKEN } = process.env;
 
 if (!ESLINT_GITHUB_TOKEN) {
-	throw new Error("Missing ESLINT_GITHUB_TOKEN.");
+	// throw new Error("Missing ESLINT_GITHUB_TOKEN.");
```
nzakas, line 48: Oops?

tools/fetch-sponsors.js
```diff
@@ -414,28 +414,29 @@ async function fetchThanksDevData() {
...
-					tier: getTierSlug(lastPayment.amount),
-				});
-			}
-
-			return accumulator;
-		},
-		[],
-	);
+			const paymentYear = new Date(lastPayment.month).getFullYear();
+			const currentYear = new Date().getFullYear();
```
nzakas, line 425: We can simplify this a bit:

```js
const now = new Date();
const isCurrentYear = now.getFullYear().toString() === lastPayment.month.slice(0, 4);
const isCurrentMonth = (now.getMonth() + 1).toString().padStart(2, "0") === lastPayment.month.slice(5, 7);
```

tools/fetch-sponsors.js
harish-sethuraman: my bad removed the comment

tools/fetch-sponsors.js
harish-sethuraman: changed

tools/fetch-sponsors.js
```diff
@@ -387,6 +391,59 @@ async function fetchGitHubSponsors() {
...
+
+			return isCurrentMonth && isCurrentYear;
+		})
+		.map(({ name, login, avatar, url, payments }) => {
+			const { amount } = payments.at(-1);
+			return {
+				name: name ?? login,
+				url: fixUrl(url),
+				image: avatar,
+				monthlyDonation: Number(amount),
```
nzakas, line 436: We can also calculate `totalDonations` by adding up the payments.

```js
totalDonations: payments.reduce((total, payment) => total + payment.amount, 0);
```

harish-sethuraman: Isnt this only for the month? Or do we want to show the total the user has paid?

amareshsm: It should be for the month, I suppose. We won’t be able to show it to the user since we’re only checking their last payment.

nzakas: The sponsors from GitHub and Open Collective all include `totalDonations` along with `monthlyDonation`. I think we should do the same here.

harish-sethuraman: done added total donations

nzakas (approved):
LGTM. Thanks!

Issue #612, opened 2024-08-12
Suggestion: Add last updated date on blog posts 


amareshsm:
### What problem do you want to solve?

Currently, our blogs do not display the last updated date on the page.

### What do you think is the correct solution?

It would enhance the user experience if the last updated date of the blog post were displayed somewhere on the page, allowing users to quickly see when it was last updated, rather than having to scroll down to the bottom of the page to check the update history.

![image](https://github.com/user-attachments/assets/bb382d4e-5d4e-42e7-b0a2-415971ebe35e)
![image](https://github.com/user-attachments/assets/83869012-ee58-4a0d-b3e8-74e5c650e180)

### Participation

- [X] I am willing to submit a pull request for this change.

### Additional comments

We could display the last updated date in the hero section or possibly on the sidebar.

![image](https://github.com/user-attachments/assets/6b273575-80b2-4451-b1fd-7db0c09d8d71)

or

![image](https://github.com/user-attachments/assets/9e9b1e65-a48b-47a8-b483-d14006c3bf80)

nzakas:
I like this idea. Let's make sure that the updated dates match the publish dates.

Also, I think the updated date should maybe go right under the published date? I'm not a fan of it going under the teaser, but open to other suggestions.

I do like it on the sidebar. :+1:

harish-sethuraman:
We do have some date visible on the blog sections. We might want to check on that? That is possibly the created date of the blog. Should we change that to updated date as well? 

+1 to keeping the updated date in the sidebar

<img width="942" alt="Screenshot 2024-08-13 at 10 24 32 AM" src="https://github.com/user-attachments/assets/af21b546-1463-46c3-9ed6-4cf32ac373a8">

nzakas:
I think we should keep the created date on the index pages. It would be confusing to see the dates out of order.

PR #692, opened 2025-02-08
feat: add last updated date to blog posts

Pixel998:
#### Prerequisites checklist

The purpose of the pull request is to add a "last updated" date to blog posts.

I added a "last updated" date to each blog post.

localhost_2022_blog_2025_01_eslint-v9 18 0-released_

#### Related Issues

Fix [#612](https://github.com/eslint/eslint.org/issues/612)


.eleventy.js
```diff
@@ -22,6 +22,7 @@ const path = require("node:path");
 const fs = require("node:fs");
 const yaml = require("js-yaml");
 const { DateTime } = require("luxon");
+const { execSync } = require("node:child_process");
```
nzakas, line 25: Is there not some way built into Eleventy to get the updated date/time? This feels a bit heavy to me.

nzakas (commented):
Thanks for taking a look at this.

Can you move the updated date/time to the left column? It doesn't look very good as another line at the top.

nzakas:
Yes, that looks good. :+1:

Tanujkanti4441 (commented):
Screenshot 2025-02-16 120259

there is no border between the author's detail and last updated date section for small screen size can you fix that?

Also proposal in https://github.com/eslint/eslint.org/issues/612 says
> It would enhance the user experience if the last updated date of the blog post were displayed somewhere on the page, allowing users to quickly see when it was last updated, rather than having to scroll down to the bottom of the page to check the update history.

and for small screen sizes we still have to scroll down to the bottom of the page, should we also add the this date in header for small screens?

nzakas:
> and for small screen sizes we still have to scroll down to the bottom of the page, should we also add this date in header for small screens?

@Tanujkanti4441 no, I don't think that necessary.

@Pixel998 I like @amareshsm's design. He's on our website team so he's a good resource to work with.

src/_data/blogDates.json
nzakas, line 1: Please rename to `blog-dates.json` to follow file naming conventions.

tools/update-blog-dates.js
```diff
@@ -0,0 +1,132 @@
+"use strict";
+
+/**
+ * @fileoverview
+ * Updates blogDates.json with last modified dates for blog posts.
+ * Processes staged blog files by default. Run with --init to process all files
+ * and initialize the dates file.
+ */
```
nzakas, line 8: Please move this comment above `"use strict"`

tools/update-blog-dates.js
```diff
@@ -0,0 +1,132 @@
...
+	return files
+		.filter(file => file.endsWith(".md"))
+		.map(file => path.resolve(BLOG_DIR, file));
+}
+
+/**
+ * Gets the list of staged blog post files from git.
+ * @returns {Promise<string[]>} Array of absolute paths to staged .md files in the blog directory.
+ */
+async function getStagedBlogFiles() {
```
nzakas, line 46: Because we're running this with lint-staged, the staged files will be passed on the command line. You can get them using `process.argv.slice(2)`.

tools/update-blog-dates.js
```diff
@@ -40,23 +40,19 @@ async function getAllBlogFiles() {
...
-		.map(file => path.resolve(process.cwd(), file));
+	const stagedFiles = process.argv.slice(2);
+
+	return stagedFiles.filter(file => {
+		const relativePath = path.relative(process.cwd(), file);
+		return (
+			relativePath.startsWith(path.normalize("src/content/blog/")) &&
+			relativePath.endsWith(".md")
+		);
+	});
```
nzakas, line 55: It's safer to use `path` methods to determine if a file is a blog post. Something like this:

```suggestion
    const blogDir = path.resolve("src/content/blog");

	return stagedFiles.filter(file => {
		const dirPath = path.dirname(file);
		
		return (
			dirPath.startsWith(blogPath)) &&
			file.endsWith(".md")
		);
	});
```

package.json
```diff
@@ -40,7 +40,8 @@
     "lint:fix:scss": "stylelint \"**/*.scss\" --fix",
     "fmt": "prettier --write .",
     "fmt:check": "prettier --check .",
-    "start": "npm-run-all build:sass --parallel *:*:watch"
+    "start": "npm-run-all build:sass --parallel *:*:watch",
+    "init-blog-dates": "node tools/update-blog-dates.js --init"
```
nzakas, line 44: To follow our conventions:
```suggestion
    "build:init-blog-dates": "node tools/update-blog-dates.js --init"
```

nzakas (approved):
LGTM. Just left one comment about script naming but otherwise seems ready.

Would like another review before merging.

src/_data/blog-dates.json
```diff
@@ -0,0 +1,432 @@
...
+    "2014-01-19-breaking-change-formatter.md": "2022-02-05T06:13:45.000Z",
+    "2014-01-20-breaking-change-config-file.md": "2022-02-05T06:13:45.000Z",
+    "2014-01-20-eslint-0.3.0-released.md": "2022-02-05T06:13:45.000Z",
+    "2014-02-12-eslint-0.4.0-released.md": "2022-02-05T06:13:45.000Z",
+    "2014-03-03-eslint-0.4.2-released.md": "2022-02-05T06:13:45.000Z",
+    "2014-03-18-eslint-0.4.3-released.md": "2022-02-05T06:13:45.000Z",
+    "2014-03-25-eslint-0.4.4-released.md": "2022-02-05T06:13:45.000Z",
+    "2014-03-29-eslint-0.4.5-released.md": "2022-02-05T06:13:45.000Z",
+    "2014-04-10-eslint-0.5.0-released.md": "2022-02-05T06:13:45.000Z",
+    "2014-04-17-eslint-0.5.1-released.md": "2022-02-05T06:13:45.000Z",
```
mdjermanovic, line 11: Looks like blog posts copied over from the old website repo (https://github.com/eslint/archive-website) will have the update date `2022-02-05`, which I think would be misleading. Perhaps we could get the last updated date from the old website repo, or maybe just use the publish date as the last updated date for these posts?

nzakas: Good catch. :+1: to using the publish date as the last updated date.

mdjermanovic: I see, there was a commit https://github.com/eslint/archive-website/commit/5aa1575a37e3b4a381b7f5f57acf31f59ee45947 that fixed the authors on that date.

mdjermanovic: Now, I think this raises a question: Does any change to the blog post's source file constitute an update worth showing in the blog post? I think readers would be interested in factual changes to the content, not e.g.,  authors, labels, etc.., maybe not even fixing small typos in the text. Since we rarely update the content of blog posts, I'd be in favor of manually updating this date rather than automatically in pre-commit.

nzakas: The problem with manual updates is that they tend not to happen. To me, the fact that we rarely update blog posts is a good argument for automatically updating the "last updated date" because we're more likely to forget.

nzakas: @mdjermanovic what do you think?

nzakas:
Ping @mdjermanovic

src/_data/blog-dates.json
mdjermanovic: Okay, we can always undo the date change directly in the PR if we don't want to modify the date based on the kind of change in the blog post file.

.eleventy.js
```diff
@@ -160,6 +160,13 @@ module.exports = eleventyConfig => {
 		value1.concat(value2),
 	);
 
+	eleventyConfig.addFilter("lastUpdated", filepath => {
+		const relativePath = path.relative("src/content/blog", filepath);
+		const blogDates = require("./src/_data/blog-dates.json");
+
+		return new Date(blogDates[relativePath]);
```
mdjermanovic, line 167: When the entry for a blog post doesn't exist in `blog-dates.json`, this results in: 

image

Perhaps we should ensure that in the mentioned case the page shows the publish date here?

nzakas (approved):
LGTM. Would like @mdjermanovic to verify before merging.

PR #698, opened 2025-02-28
feat: add copy button

xbinaryx:
#### Prerequisites checklist

While reading the ESLint CSS support post, I noticed the code blocks lacked a copy button. I had to select and copy the text manually.

I added a copy button to code blocks.


Tanujkanti4441 (commented):
Hi @xbinaryx, thanks for the PR.

Right now copy button appears on hover as also suggested in [this comment we](https://github.com/eslint/eslint/pull/18855#issuecomment-2334358659) can make it appear always both button and its background.

Also it overrides the content behind it like in these examples
Screenshot 2025-02-28 185416
or
Screenshot 2025-02-28 185434

should we place it border as we do with correct/incorrect logo like this

Screenshot 2025-02-28 185834

Would like more opinions on this from team.

Tanujkanti4441:
Hi @xbinaryx, for now can you make the copy button appear all the time with its background like github code block?

Screenshot 2025-02-28 185434

we can later change it's position if overriding code will be an issue.

src/assets/scss/blog.scss
```diff
@@ -153,3 +153,63 @@ ul.tag-list {
...
+	align-items: start;
+	justify-content: space-between;
+	margin: 1.5rem 0;
+	border-radius: var(--border-radius);
+	background-color: var(--lightest-background-color);
+
+	> pre {
+		margin: 0;
+	}
+}
```
Tanujkanti4441, line 168: Screenshot 2025-03-11 180544

I not sure if this is the right approach, we can just add some padding (`padding-right: 3.5rem`) to `pre` tag
https://github.com/eslint/eslint.org/blob/1b26995f615351e3f969c76b5bbc2486a6f56ab5/src/assets/scss/syntax-highlighter.scss#L30-L37

so can you please revert this change.

and also i think we can keep the background always not just on hover to make it more visible. 
Screenshot 2025-02-28 185434

Tanujkanti4441 (changes requested):

src/assets/scss/blog.scss
nzakas: Agree with @Tanujkanti4441. The background color makes it a lot more visible. :+1:

src/_11ty/plugins/preWrapper.js
nzakas, line 1: Filenames should be in dash-case, so please rename to `pre-wrapper.js`

Tanujkanti4441 (commented):
Screenshot 2025-03-13 172600

The contrast is very less between copy button and code slate's background specially in light mode.

how about adding a border to the copy button for both themes. something like this?

Screenshot 2025-03-13 172534
Screenshot 2025-03-13 172459

Tanujkanti4441 (approved):
LGTM! thanks for contributing.

Would like another team member to review this.

nzakas (approved):
This looks fantastic, great work!

PR #713, opened 2025-03-29
refactor: copy to clipboard code

amareshsm:
#### Prerequisites checklist

Refactored Copy to Clipboard functionality:
1. Cloning and then copying the `textContent` seems unnecessary; it's more efficient to access the `textContent` directly.
2. Assigning the same ID to all buttons is not recommended, as IDs should be unique. It's better to use a classname instead.
3. Use .then() with error handling when working with the navigator clipboard API because clipboard operations are asynchronous and can fail for several reasons like permission issues and Clipboard API Access Restrictions.


nzakas (approved):
LGTM. Thanks!

PR #724, opened 2025-05-04
feat: add report an issue button

SwetaTanwar:
#### Prerequisites checklist

- Add `Report an issue` button under Share section on playground

- Prefilled `Environment` with ESLint Playground text
- Prefilled `What did you do` with code, config and output
- Prefilled `What actually happened?` with output
- Prefilled `Link to Minimal Reproducible Example` with the shareable URL

#### Related Issues
Fix for https://github.com/eslint/eslint.org/issues/598

Need opinion for the following edge case

- There is an edge case where if you enable add the rules, the config file becomes too big, and thus trying to create issue gives error of request URL too long. 

Following are the identified issues in case when all the rules were enabled

1. The shareable link which is generated (base64 config) is pretty big and we append it in the params, which when gets URL encoded becomes too big
2. We are adding template message only for the description field, but no such thing(length check) is being done for error output which in case of being too big causes error. 

Few probable solutions as per me could be

1. Shareable link will be always added irrespective of its length. but need to figure out if we can somehow shorten it
2. In case of too large error output, we can skip adding it by default to the error output field, since in description field as well, we add the error output.

#### Screenshots

Prefilled `Environment` 
Screenshot 2025-05-04 at 10 27 53 PM

Prefilled `What did you do` 
Screenshot 2025-05-04 at 10 29 48 PM

Prefilled `What actually happened?` with error output and `Link to Minimal Reproducible Example` with the shareable URL 
Screenshot 2025-05-04 at 10 30 12 PM

Prefilled `What did you do` with template message when config is too big 
Screenshot 2025-05-05 at 8 42 22 AM

Pasted config content in `What did you do` from clipboard when config is too big 
Screenshot 2025-05-05 at 8 42 39 AM

### DIFF
--- a/src/assets/scss/components/playground-configuration.scss
+++ b/src/assets/scss/components/playground-configuration.scss
@@ -246,32 +246,6 @@
 	margin-top: 1rem;
 }
 
-.report-issue-btn {
-	display: inline-block;
-	background-color: var(--color-neutral-100);
-	border: 1px solid var(--border-color);
-	color: var(--body-text-color);
-	padding: 0.5rem 1rem;
-	border-radius: var(--border-radius);
-	font-size: 0.875rem;
-	cursor: pointer;
-	transition: all 0.2s ease;
-
-	&:hover {
-		background-color: var(--color-neutral-200);
-		color: var(--link-color);
-	}
-
-	[data-theme="dark"] & {
-		background-color: var(--color-neutral-700);
-		color: var(--color-neutral-100);
-
-		&:hover {
-			background-color: var(--color-neutral-600);
-		}
-	}
-}
-
 .playground__config-options__section--rules {
 	font-size: 0.875rem;
 }
--- a/src/assets/scss/components/playground-configuration.scss
+++ b/src/assets/scss/components/playground-configuration.scss
@@ -242,6 +242,36 @@
 	}
 }
 
+.report-issue-wrapper {
+	margin-top: 1rem;
+}
+
+.report-issue-btn {
+	display: inline-block;
+	background-color: var(--color-neutral-100);
+	border: 1px solid var(--border-color);
+	color: var(--body-text-color);
+	padding: 0.5rem 1rem;
+	border-radius: var(--border-radius);
+	font-size: 0.875rem;
+	cursor: pointer;
+	transition: all 0.2s ease;
+
+	&:hover {
+		background-color: var(--color-neutral-200);
+		color: var(--link-color);
+	}
+
+	[data-theme="dark"] & {
+		background-color: var(--color-neutral-700);
+		color: var(--color-neutral-100);
+
+		&:hover {
+			background-color: var(--color-neutral-600);
+		}
+	}
+}
+
 .playground__config-options__section--rules {
 	font-size: 0.875rem;
 }
--- a/src/playground/components/ShareURL.js
+++ b/src/playground/components/ShareURL.js
@@ -19,7 +19,9 @@ const buildGitHubIssueDescription = (code, config, errorOutput) => {
 	}
 
 	if (config) {
-		parts.push(`### ESLint Configuration\n${formatCodeBlock(config)}\n`);
+		parts.push(
+			`<details> \n <summary> ESLint Configuration </summary> \n\n${formatCodeBlock(config)}\n </details>\n`,
+		);
 	}
 
 	if (errorOutput) {
--- a/src/playground/components/ShareURL.js
+++ b/src/playground/components/ShareURL.js
@@ -3,6 +3,8 @@ import {
 	GITHUB_ISSUE_URL,
 	MAX_URL_LENGTH,
 	CLIPBOARD_FALLBACK_MESSAGE,
+	LINT_OUTPUT_FALLBACK_MESSAGE,
+	REPRO_URL_FALLBACK_MESSAGE,
 } from "../utils/constants";
 
 // Helper function for code template formatting
@@ -20,7 +22,7 @@ const buildGitHubIssueDescription = (code, config, errorOutput) => {
 
 	if (config) {
 		parts.push(
-			`<details> \n <summary> ESLint Configuration </summary> \n\n${formatCodeBlock(config)}\n </details>\n`,
+			`### ESLint Configuration \n<details> \n <summary> Tap to see full config </summary> \n\n${formatCodeBlock(config)}\n </details>\n`,
 		);
 	}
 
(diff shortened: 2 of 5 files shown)


src/playground/components/ShareURL.js
```diff
@@ -89,6 +202,14 @@ export default function ShareURL({ url }) {
...
 						)}
 					</div>
+					<div className="report-issue-wrapper">
+						<button
+							className="c-btn report-issue-btn"
+							onClick={handleReportIssue}
+						>
+							Report an issue
+						</button>
+					</div>
```
snitin315, line 212: Let's use a secondary style button as seen in the screenshot attached here https://github.com/eslint/eslint.org/issues/598#issuecomment-2287875985

Right now it doesn't match with button styles through the site.

Screenshot 2025-05-06 at 8 30 28 AM

src/playground/components/ShareURL.js
```diff
@@ -1,16 +1,141 @@
...
+const buildGitHubIssueDescription = (code, config, errorOutput) => {
+	const parts = [];
+
+	if (code) {
+		parts.push(`### Playground Code\n${formatCodeBlock(code)}\n`);
+	}
+
+	if (config) {
+		parts.push(`### ESLint Configuration\n${formatCodeBlock(config)}\n`);
+	}
```
snitin315, line 23: Since the config can be long, let's use a dropdown here:

```suggestion
	if (config) {
		parts.push(`<details> \n <summary> ESLint Configuration </summary> \n${formatCodeBlock(config)}\n </details>`);
	}
```

https://github.com/user-attachments/assets/3fd3a4f5-750d-4e83-a83d-520dc3317417

src/playground/components/ShareURL.js
```diff
@@ -1,16 +1,141 @@
...
+} from "../utils/constants";
 
-export default function ShareURL({ url }) {
+// Helper function for code template formatting
+const formatCodeBlock = (code, lang = "js") => `\`\`\`${lang}
+${code}
+\`\`\``;
+
+// Helper to build GitHub issue description
+const buildGitHubIssueDescription = (code, config, errorOutput) => {
```
snitin315, line 14: Let's also add a default title similar to typescript-eslint:  `Bug: [rule name] <write short description here>`

https://github.com/typescript-eslint/typescript-eslint/blob/d5209cd686f1c74f854c3d1e9624fe744ed170e9/packages/website/src/components/lib/markdown.ts#L70

snitin315 (changes requested):
Thanks for picking this up. Left a few suggestions.

>There is an edge case where if you enable add the rules, the config file becomes too big, and thus trying to create issue gives error of request URL too long.

I think it's okay to skip handling this case for now, since reproductions are meant to be minimal or focused on a specific rule—it's quite rare that someone would need to enable all rules for a repro.

src/assets/scss/components/playground-configuration.scss
```diff
@@ -242,6 +242,36 @@
...
+}
+
+.report-issue-btn {
+	display: inline-block;
+	background-color: var(--color-neutral-100);
+	border: 1px solid var(--border-color);
+	color: var(--body-text-color);
+	padding: 0.5rem 1rem;
+	border-radius: var(--border-radius);
+	font-size: 0.875rem;
```
lumirlumir, line 256:
```suggestion
	font-size: 1rem;
```

Thanks for opening a PR!

How about setting the `font-size` to `1rem`, since the other buttons in the Playground also use `1rem`?

Here are the other buttons:

image

image

src/playground/components/ShareURL.js
SwetaTanwar: Fixed

src/playground/components/ShareURL.js
SwetaTanwar: Fixed

src/assets/scss/components/playground-configuration.scss
SwetaTanwar: Used the styling for secondary-btn class

src/playground/components/ShareURL.js
SwetaTanwar: Fixed

Screenshot 2025-05-10 at 5 03 04 PM

snitin315 (dismissed):
LGTM, thank you for the contribution ❤️ 

Leaving ti open for @lumirlumir

amareshsm (commented):
If we apply formatting to the rules and the number of rules is high (e.g., it breaks when more than 53 rules are added), an error occurs when redirecting to GitHub to create a new issue.

image


The same feature has already been attempted in [this PR](https://github.com/eslint/eslint.org/pull/616). Please check the PR and ensure any related cases are properly handled.

lumirlumir:
I'll review it again once the above suggestion is addressed!

snitin315:
@SwetaTanwar Let's just show an error in case the URL is very long , also discussed in  https://github.com/eslint/eslint.org/pull/616#issuecomment-2321462844

SwetaTanwar:
The issue arises from the combined length of the link, config, and error — together, they exceed the URL character limit and break the link.

The proposed solution of copying the config to the clipboard only helps by removing it from the URL, but even then, the link and error alone can still exceed the 8148 character limit.

What do you suggest we do in this case?

One option is to remove the error from the URL, since the link must remain intact. However, this would leave the error field empty, and users would then need to manually copy and paste the error object.

Or altogether show the error for when URL is very long?
Also what should be the error message?

SwetaTanwar:
As per this [comment](https://github.com/eslint/eslint.org/pull/616#issuecomment-2344056213), the issue is being created in blank issue template, which does not exist as of now in eslint repo. We should have one.

snitin315:
The case seems to be handled now @lumirlumir @amareshsm 

https://github.com/user-attachments/assets/9da45c0c-b069-4f8f-8ee0-07e544559ad2

lumirlumir:
I've found another bug.

When entering the Playground for the first time, the "Report an issue" button correctly includes the Playground code.

However, if I change a setting like the ECMAScript version, the code is no longer pasted correctly.

Here's a short video that reproduces the issue:

https://github.com/user-attachments/assets/bfeda7b9-e9a6-4dcb-af61-30ce18392037

SwetaTanwar:
> I've found another bug.
> 
> When entering the Playground for the first time, the "Report an issue" button correctly includes the Playground code.
> 
> However, if I change a setting like the ECMAScript version, the code is no longer pasted correctly.
> 
> Here's a short video that reproduces the issue:
> 
>  _2025_05_12_19_06_54_829.mp4

I will take a look at this.

SwetaTanwar:
@snitin315 @nzakas @lumirlumir Can I get a review?

lumirlumir (approved):
@SwetaTanwar 

Thanks for your detailed explanation. It works well when the limit is not exceeded.

LGTM. Would like @snitin315 and @amareshsm to review before merging.

snitin315 (approved):
LGTM, thanks!

nzakas (approved):
LGTM. Thanks!

PR #744, opened 2025-07-20
feat: add condition to check TS parser

Tanujkanti4441:
#### Prerequisites checklist

https://github.com/eslint/eslint/blob/f46fc6c137c951bc73cf3bd9446053c1b11f769b/docs/.eleventy.js#L215-L217

To set the parser option to `@typescript-eslint/parser` in playground on `languageOptions.parser = typeScriptESLintParser` option, this will enable the TS parser when someone directly open the playground from `open in playground` button in docs.

Added a condition to check if `parser === typeScriptESLintParser` then set the playground parser to `@typescript-eslint/parser`.

Need to add this change before adding playground button for TypeScript code example https://github.com/eslint/eslint/pull/19671

### DIFF
--- a/src/playground/components/Configuration.js
+++ b/src/playground/components/Configuration.js
@@ -108,18 +108,6 @@ const defaultOption = {
 
 const isEmpty = obj => Object.keys(obj).length === 0;
 
-const parserValue = parser => {
-	if (
-		(typeof parser === "object" &&
-			parser.meta.name === "typescript-eslint/parser") ||
-		parser === "@typescript-eslint/parser"
-	) {
-		return "@typescript-eslint/parser";
-	}
-
-	return null;
-};
-
 export default function Configuration({
 	initialOptions,
 	rulesMeta,
@@ -131,6 +119,7 @@ export default function Configuration({
 	validationError,
 	rulesWithInvalidConfigs,
 	setRulesWithInvalidConfigs,
+	typeScriptESLintParser,
 }) {
 	const [showVersion, setShowVersions] = useState(false);
 	const [showRules, setShowRules] = useState(true);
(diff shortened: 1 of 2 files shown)


lumirlumir (approved):
LGTM, thanks!

Would like another review before merging.

src/playground/components/Configuration.js
```diff
@@ -272,6 +261,19 @@ export default function Configuration({
 		return config;
 	};
 
+	const parserValue = parser => {
+		if (
+			(typeof parser === "object" &&
+				parser.meta.name === "typescript-eslint/parser") ||
+			parser === "@typescript-eslint/parser" ||
+			parser === typeScriptESLintParser
```
mdjermanovic, line 269: I'm not sure I understand this change. Shouldn't the `Open in Playground` button send `"parser":"@typescript-eslint/parser"` in `options.languageOptions`, not the parser object?

Tanujkanti4441: I tried `"parser":"@typescript-eslint/parser"` but got this error, and it is not working as expected:

image

mdjermanovic: Can you share a branch with a code change that is causing this error?

Note that we want to pass `"parser":"@typescript-eslint/parser"` to the playground, not to the code that underlines errors in code example on the docs site.

https://github.com/eslint/eslint/blob/f46fc6c137c951bc73cf3bd9446053c1b11f769b/docs/.eleventy.js#L231-L236

Tanujkanti4441: Seems you are right, `"parser":"@typescript-eslint/parser"` is working and have implemented in https://github.com/eslint/eslint/pull/19671.

So I think this PR can be closed, thanks.

Issue #910, opened 2026-01-16
Bug: Playground ignores sourceType when using @typescript-eslint/parser


xbinaryx:
### URL(s)

[...]

PR #915, opened 2026-01-19
fix: remove special handling of sourceType for TypeScript parser

### LINKED ISSUE #891: Suggestion: Move `languageOptions` properties from nested `parserOptions`
### What problem do you want to solve?

Currently the playground generates a configuration which specifies `sourceType` and `ecmaVersion` in `languageOptions.parserOptions` instead of `languageOptions`. For example:
```javascript
export default [
    {
        "languageOptions": {
            "parserOptions": {
                "ecmaFeatures": {
                    "jsx": true
                },
                "sourceType": "module"
            },
        }
    }
];
```

### What do you think is the correct solution?

Move the properties `languageOptions.parserOptions.{sourceType,ecmaVersion}` to `languageOptions.{sourceType,ecmaVersion}`.

### Participation

- [x] I am willing to submit a pull request for this change.

### Additional comments

This change would be consistent with the [documentation](https://eslint.org/docs/latest/use/configure/language-options#specifying-javascript-options)

DMartens:
#### Prerequisites checklist

Removing the special behaviour for the TypeScript parser.

This PR includes two fixes:
1. remove the hardcoded `sourceType` of "module" for the TypeScript parser (refs #910)
2. Fixes #891 by setting the `sourceType` on `languageOptions` instead of `languageOptions.parserOptions` when swapping the parser.

#### Related Issues
Fixes #891 and #910.

Please also try various variations of updating the configuration and the resulting configuration file.

### DIFF
--- a/src/playground/app.jsx
+++ b/src/playground/app.jsx
@@ -45,8 +45,19 @@ const fillOptionsDefaults = options => ({
 });
 
 const convertLegacyOptionsToFlatConfig = options => {
-	// If there are no legacy properties, return it
-	if (!options.env && !options.parserOptions && !options.languageOptions) {
+	const isFlatConfig = !options.env && !options.parserOptions;
+	if (isFlatConfig) {
+		/**
+		 * Move `sourceType` from `languageOptions.parserOptions` to `languageOptions`.
+		 * This is a fix for previous links using the `@typescript-eslint/parser`.
+		 * @see https://github.com/eslint/eslint.org/issues/891
+		 */
+		if (options.languageOptions?.parserOptions?.sourceType !== void 0) {
+			options.languageOptions.sourceType =
+				options.languageOptions.parserOptions.sourceType;
+			delete options.languageOptions.parserOptions.sourceType;
+		}
+
 		return options;
 	}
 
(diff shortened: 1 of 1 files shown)


src/playground/app.jsx
Tanujkanti4441: Screenshot 2026-01-20 225100

The `Source Type` dropdown is not selecting any option, even after clicking on the option. Can you fix that?

src/playground/components/configuration.jsx
```diff
@@ -532,8 +532,6 @@ export default function Configuration({
 											selected.value;
 										delete newOptions.languageOptions
 											.sourceType;
-										newOptions.languageOptions.parserOptions.sourceType =
-											"module";
```
Tanujkanti4441: Currently, we are setting `Source Type` to `module` when selecting `@typescript-eslint/parser` then user can select any `Source Type` they want. Can we retain this behavior?

Tanujkanti4441 (changes requested):
Thanks for the PR.

Tanujkanti4441 (commented):
Image

Right now on selecting `@typescript-eslint/paser` the `Source Type` option is empty, as  https://github.com/eslint/eslint.org/pull/915#discussion_r2725504081 suggests to use the default value for `sourceType`, so can we also add `(default)` option on `Source Type` for tsparser as well like we have for espree?

src/playground/components/configuration.jsx
```diff
@@ -391,28 +385,13 @@ export default function Configuration({
...
-										? sourceTypeOptionsForTypeScriptParser
-										: sourceTypeOptions
-								}
+								value={sourceTypeOptions.filter(
+									sourceTypeOption =>
+										sourceTypeOption.value ===
+										(options.languageOptions?.sourceType ||
+											"default"),
+								)}
+								options={sourceTypeOptions}
```
Tanujkanti4441: Getting this error in playground.

Image

steps to reproduce
1. Select the `@typescript-eslint/parser`, the sourceType will be (default) here.
2. Then select any other source type like `script`.
3. Select the (default) again.

Note:- I tried to run the changes in this PR locally, as I am not able to see the playground using the deploy link.

PR #1004, opened 2026-04-16
fix: correctly guard "Fix All" when no fixable messages

amareshsm:
#### Prerequisites checklist

#### AI acknowledgment

Fixes a logic bug in the playground's "Fix All" action so it correctly detects when there are no auto-fixable ESLint messages and returns early. Previously the guard was ineffective, causing an unnecessary call to `linter.verifyAndFix()` on every click even when nothing was fixable.

Updated app.jsx: replaced `filter` with `some`

Before
```
const hasFixableMessages = messages.filter(message => message.fix)
 ```
 
 
After: 
```
const hasFixableMessages = messages.some(message => message.fix)
```

### DIFF
--- a/src/playground/app.jsx
+++ b/src/playground/app.jsx
@@ -252,7 +252,7 @@ const App = () => {
 
 	const onFixAll = () => {
 		const hasFixableMessages = messages.filter(message => message.fix);
-		if (!hasFixableMessages) {
+		if (!hasFixableMessages.length) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/playground/app.jsx [hunk 1]
```diff
@@ -252,7 +252,7 @@ const App = () => {
 
 	const onFixAll = () => {
 		const hasFixableMessages = messages.filter(message => message.fix);
```
DMartens: As this is a boolish context, I think we rather should use `.some` instead of `.filter` (and not use the `.length` property).

amareshsm: Update 👍 .

DMartens (approved):
Changes LGTM, thanks.

PR #1011, opened 2026-04-20
fix: use `find()` for single-value playground selects

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

This PR aligns the playground configuration selects with the `react-select` API for single-value selects. The existing code used `.filter()` for some single-select `value` props, which returns an array and is better suited to `isMulti` selects.

Updated the single-value `react-select` fields to use `.find()` instead of `.filter()`.


DMartens (approved):
Changes LGTM, thanks.

PR #1029, opened 2026-05-13
fix: prevent duplicate listeners and implicit global var in nav script

amareshsm:
#### Prerequisites checklist

#### AI acknowledgment

Fix a couple of small bugs and clean up some code in the nav scripts:

1. After the user opens the mobile nav, resizes to desktop, then resizes back to mobile, the internal open flag drifts out of sync with the DOM's data-open attribute. The next click on the toggle then closes the (already-closed) menu instead of opening it, so the user has to click twice.

2. Three variables (btn_theme, currentButton, url) were being assigned without a declaration keyword. Because the scripts aren't in strict mode, those assignments leak onto window as implicit globals.

3. A leftover body variable was declared but never used.

1. Removed the unused body variable in main.js and components-index.js.

2. In WidthChange / initIndex, reset the open flag to match the data-open attribute whenever the viewport crosses the breakpoint, so the toggle behaves correctly after resizing.

3. Added missing const declarations for btn_theme, currentButton, and url to prevent implicit globals.

4. Defensive checks - Added a null guard in unsetOtherButtons in case no button has aria-pressed="true" when it's called and Added removeEventListener calls before re-attaching the click handler

### DIFF
--- a/src/assets/js/components-index.js
+++ b/src/assets/js/components-index.js
@@ -1,7 +1,6 @@
 (function () {
 	var index_trigger = document.getElementById("js-index-toggle"),
 		index = document.getElementById("js-index-list"),
-		body = document.getElementsByTagName("body")[0],
 		open = false;
 
 	if (matchMedia) {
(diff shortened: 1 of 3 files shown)


DMartens (approved):
Changes LGTM, thanks.

PR #1045, opened 2026-05-29
fix: border radius of suggestion alert

Tanujkanti4441:
#### Prerequisites checklist

#### AI acknowledgment

To fix the border radius of the suggestion alert, when there is only one suggestion, as right now, the suggestion alert is looking like this when hovering (no border radius in the bottom left):
Screenshot 2026-05-29 224808
Screenshot 2026-05-29 154927

Applied border radius to the alert when there is only one suggestion.


amareshsm (approved):
LGTM

PR #1057, opened 2026-06-11
feat: show fix all and disable all only with multiple messages

Tanujkanti4441:
#### Prerequisites checklist

#### AI acknowledgment

Right now, the `Disable All` and `Fix All` buttons appear even if there is only one `Disable` or `Fix` button in the error message in that case, we don't need `Disable All` or `Fix All` to only disable one rule or fix one issue.

Screenshot 2026-06-11 183603

So, this pr gives a fix for that.

Made `Disable All` and `Fix All` buttons appear only when there are more than one messages with `Disable` or `Fix` buttons.

Screenshot 2026-06-11 191220
Screenshot 2026-06-11 191821
Screenshot 2026-06-11 191710

### DIFF
--- a/src/playground/app.jsx
+++ b/src/playground/app.jsx
@@ -314,10 +314,10 @@ const App = () => {
 		new Set([]),
 	);
 
-	const hasMultipleFixMessages = messages.filter(message => message.fix).length > 1;
-	const hasMultipleDisableMessages = messages.filter(
-		message => options.rules[message.ruleId],
-	).length > 1;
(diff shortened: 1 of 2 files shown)


DMartens (dismissed):
Changes LGTM, thanks. Leaving open for second review as it is a behavior change.

Pixel998:
@Tanujkanti4441 Could you take a look at the CI failure? It looks like there's a formatting error.

mdjermanovic (approved):
LGTM, thanks!

PR #1064, opened 2026-06-16
fix: Add missing EOF in all_authors.json

amirabbas-gh:
#### Prerequisites checklist

- [ ] I have read the [contributing guidelines](https://github.com/eslint/.github/blob/master/CONTRIBUTING.md).

#### AI acknowledgment

- [ ] I did _not_ use AI to generate this PR.


mdjermanovic:
Thanks for the PR, but this file is autogenerated so we'd need to update all places that are modifying it, while it doesn't seem that missing EOF is causing any problems so I think it's fine to leave it as is.

PR #1069, opened 2026-06-21
fix: prevent playground scrolling when deprecation notice is present

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

This PR fixes an issue where the Playground page becomes scrollable whenever the deprecation notice is present at the top of the page. Previously, the height of the playground layout was hardcoded to `calc(100vh - 77px)`. While this subtracted the height of the header, it failed to account for the deprecation notice. When the notice appeared, it pushed the playground down, causing the `100vh` container to overflow and trigger a full-page scrollbar.

- Added a `hook: "playground_page"` to the playground page frontmatter to target the playground body.
- Refactored `playground-layout.scss` to use percentage-based heights instead of pixel deductions.

### DIFF
--- a/src/assets/scss/playground-layout.scss
+++ b/src/assets/scss/playground-layout.scss
@@ -1,4 +1,16 @@
+body.playground-page {
+	height: 100%;
+
+	#main,
+	#playground-app {
+		height: 100%;
+		min-height: 0;
+	}
+}
+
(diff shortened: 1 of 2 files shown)


DMartens (approved):
Changes LGTM, thanks.

PR #1157, opened 2026-09-11
fix: default playground theme switcher pressed state to system

Pixel998:
#### Prerequisites checklist

#### AI acknowledgment

On a first visit to the playground, none of the Light / System / Dark buttons in the footer shows as pressed. The switcher initializes its state straight from `localStorage.getItem("theme")`, which is `null` until the user clicks a button, so every button renders with `aria-pressed="false"` even though the page is already displayed in the system theme.

The component now falls back to `"system"` when no theme is stored.

### DIFF
--- a/src/playground/components/theme-switcher.jsx
+++ b/src/playground/components/theme-switcher.jsx
@@ -1,7 +1,9 @@
 import { useEffect, useState } from "react";
 
 export default function ThemeSwitcher() {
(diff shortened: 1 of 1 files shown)


DMartens (approved):
Changes LGTM, thanks.

amareshsm (approved):

